PR #197, opened 2020-06-03
leverage queueExec methodologies to support file overrides down the line.

Azaezel:
In doing so, offload the loadStartup(); from UI::initClient to the startupgui.cs file itself.
reminder notes:
utility definitions and a couple debug switches are in https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Templates/BaseGame/game/core/utility/scripts/module.cs

   %this.queueExec("/guis/mainMenu.gui"); in module A and module B will attempt to execute both
   %this.queueExec("/guis/mainMenu.gui",true) in module B with B set to depend on A will force override the dependencies script execution call.


Azaezel:
superseded by #260

PR #199, opened 2020-06-03
implement copyResource, fix copyToBmp GFXFormatR16G16B16A16F case

Azaezel:


Azaezel:
will need to retool this a lil to account for mips

rextimmy:
glCopyImageSubData is from GL 4.3 and is not available on macOS. If you want to use it anyway, you should be checking mCapabilities.copyImage first and also supply a fallback method that will work with GL 3.3 (srcTex->PBO->dstTex is one such method).

Azaezel:
Thanks for the catch. Will revisit the glTexSubImage2D route. Still not quite 100% if the no mips bit is coming form the clone or GFXGLTextureObject::copyToBmp at time of writing. Probably go ahead and upload both shortly as a pause point for the night if nothin blows up worse.

Azaezel:
https://cdn.discordapp.com/attachments/429699605572222988/717900374157295676/unknown.png
https://cdn.discordapp.com/attachments/429699605572222988/717900281479692339/unknown.png
status as of last commit.

PR #202, opened 2020-06-05
implement copyResource, fix copyToBmp

Azaezel:
GFXFormatR16G16B16A16F case wasn't accounting for mips, nor was it using the right memcpy aperture

(cleaned up and functional variant of #199)

### DIFF
--- a/Engine/source/gfx/gl/gfxGLDevice.cpp
+++ b/Engine/source/gfx/gl/gfxGLDevice.cpp
@@ -47,6 +47,7 @@
 #include "gfx/gl/gfxGLVertexAttribLocation.h"
 #include "gfx/gl/gfxGLVertexDecl.h"
 #include "shaderGen/shaderGen.h"
+#include "gfxGLUtils.h"
 
 GFXAdapter::CreateDeviceInstanceDelegate GFXGLDevice::mCreateDeviceInstance(GFXGLDevice::createInstance); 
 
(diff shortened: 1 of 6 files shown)


Azaezel:
one thing @rextimmy noted in the prior implementations first revision is that glCopyImageSubData is 4.3 spec only, so for followup, we'll want to eyeball

https://github.com/TorqueGameEngines/Torque3D/blob/92e79cf751f92dd01ce7e8ab766028bbcc37533b/Engine/source/gfx/gl/gfxGLWindowTarget.cpp#L94

https://github.com/TorqueGameEngines/Torque3D/blob/92e79cf751f92dd01ce7e8ab766028bbcc37533b/Engine/source/gfx/gl/gfxGLTextureTarget.cpp#L424

and https://github.com/TorqueGameEngines/Torque3D/blob/92e79cf751f92dd01ce7e8ab766028bbcc37533b/Engine/source/gfx/gl/gfxGLCubemap.cpp#L487

Unlike this one, who's primary usage is in the baking pipeline, some of those others are runtime sensitive, so we'll likely want a branching approach there as he'd suggested for best performance.

rextimmy:
Yeah it is totally fine to use glCopyImageSubData with gpu drivers that support it, in fact i would recommend it as it very likely leads to *the fast path* in the driver. You just need to create a fallback though like is done with the WindowTarget & TextureTarget examples above.

Azaezel:
ok, so as a status update for this weekend since this is an active blocker for opengl exclusive users, I should note it does appear there's an oversight left over at minimum with Irradiance.dds generation and mips we still need to resolve as well as that re-review for fallback purposes. if that isn't fully resolved by end of night, will likely commit this partial work so folks are at least all on the same page regarding the baseline to follow up from.

Issue #201, opened 2020-06-04
DirectX only project


Bloodknight:
When selecting DiX11 and deselecting OpenGL in CMake,  this happens.

```
2>sdlPlatformGL.cpp
2>E:\Torque3D_4.0\Engine\source\gfx/gl/tGL/tGL.h(26,10): fatal error C1083: Cannot open include file: 'glad/glad.h': No such file or directory
2>Done building project "MyGameDX.vcxproj" -- FAILED.
3>------ Skipped Build: Project: INSTALL, Configuration: Debug x64 ------
3>Project not selected to build for this solution configuration 
========== Build: 1 succeeded, 1 failed, 20 up-to-date, 1 skipped ==========
```

Azaezel:
resolved

PR #204, opened 2020-06-07
D3D-only compile fix

Areloch:
A correction to the cmake file so if OGL is disabled, it doesn't include the sdl platform GL file, allowing the project to compile.


Areloch:
For #201

Issue #205, opened 2020-06-10
IBLUtilities drawPrimitive


rextimmy:
I noticed this bug in our build the other day and you guys have the same, the primitive count here is wrong (this was my bad) https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Engine/source/T3D/lighting/IBLUtilities.cpp#L77 & here https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Engine/source/T3D/lighting/IBLUtilities.cpp#L180  Primitive count should be 1, 3 is the vertex count :)

Azaezel:
resolved

PR #206, opened 2020-06-10
adress #205

Azaezel:


Azaezel:
@rextimmy : seemed to check out this end, so if that's the mods you made, will throw that in tonight since that'll make it cross the 'checked by at least 2 folks' threshold.

rextimmy:
Yeah it will check out, you are only drawing a single triangle for this, i accidentally put in the vertex count originally not the primitive count.

PR #212, opened 2020-06-13
export rounds LODs to the nearest power of 2

Azaezel:
prevent submeshes from flickering in and out due to lod mismatches when exporting groups to a singular model file


Azaezel:
as a followup, at some point it'd be a good idea to further refine that to gen lods relative to the resulting aggregate bounds box size. this *just* handles the flicker in/out case that end users would notice.

ghost:
As an artist I always found it very unintuitive to name the LOD based on the pixel size that they will swap out, why not just name them LOD1, LOD2 etc? And then in the engine define when they are swapped out and if not, revert to some default values.

PR #213, opened 2020-06-13
typofix in opengl terrain shadergen

Azaezel:


Azaezel:
should complete resolution of #11

PR #218, opened 2020-06-13
adress #16 - don't need to swizzle vert colors

Azaezel:


Azaezel:
https://cdn.discordapp.com/attachments/580242040730943509/721420672529858630/vertColorCheck.dae verification mesh

Bloodknight:
tested and working as expected

Issue #179, opened 2020-05-30
HDR does not treat the sun as bright


lukaspj:
This issue was created in the [GarageGames Repository](https://github.com/GarageGames/Torque3D) ([Link to original issue](https://github.com/GarageGames/Torque3D/issues/4)).
The issue was originally created by @crabmusket and had a total of 1 comments that may contain additional information. The original issue description is pasted below:

**Build:** 1.2 confirmed, though I've seen it happen for as long as we've had HDR
**Platform:** Windows 7 64 bit
**Target:** PostFX
**Issue:** HDR does not treat the sun as a bright object

Something I've noticed for a while and been meaning to ask about. When you turn HDR on and start at the sun, everything gets really bright, presumably because the HDR shader doesn't register the sun as a 'bright' object. Is there anything we can do about this?

**Reproduction:** Open up Empty Terrain from the Full template, add HDR via the PostFX manager, look at the sun. Everything becomes brighter when it should become dimmer.

As described [here](http://www.garagegames.com/community/forums/viewthread/131715)

Azaezel:
RnD note: https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Templates/BaseGame/game/core/rendering/shaders/scatterSkyP.hlsl#L66

Azaezel:
resolved

Issue #116, opened 2020-05-30
Possible redundant code in scatterSkyP.hlsl


lukaspj:
This issue was created in the [GarageGames Repository](https://github.com/GarageGames/Torque3D) ([Link to original issue](https://github.com/GarageGames/Torque3D/issues/1039)).
The issue was originally created by @RichardsGameStudio and had a total of 1 comments that may contain additional information. The original issue description is pasted below:

In shaders/common/scatterSkyP.hlsl there are 2 lines of code which don't do anything.

   float4 nightSkyColor = texCUBE( nightSky, -In.v3Direction );
   nightSkyColor = lerp( nightColor, nightSkyColor, useCubemap );

   float fac = dot( normalize( In.pos ), sunDir );//<== REDUNDANT?
   fac = max( nightInterpAndExposure.y, pow( saturate( fac ), 2 ) );//<== REDUNDANT?
   Out = lerp( color, nightSkyColor, nightInterpAndExposure.y );

   Out.a = 1;

As far as I can see they can be removed, because fac isn't used anywhere.

Azaezel:
fars I can tell from
![Untitled](https://user-images.githubusercontent.com/4236451/84577350-1b61c480-ad81-11ea-958f-e15f1ce093d3.jpg)
![Untitled2](https://user-images.githubusercontent.com/4236451/84577351-1c92f180-ad81-11ea-9a41-954b1fa03f36.jpg)

fac is an attempt at tossing the value into the same colorspace

Azaezel:
resolved

PR #227, opened 2020-06-23
adress #116 and #179 (shaderside)

Azaezel:
kills off code folks haven't been using for years now, as well as a clamp to force the color into a 0-1 range. (hdr relies on >1.0 values for it's blur effect to populate)

### DIFF
--- a/Templates/BaseGame/game/core/rendering/shaders/gl/scatterSkyP.glsl
+++ b/Templates/BaseGame/game/core/rendering/shaders/gl/scatterSkyP.glsl
@@ -60,13 +60,9 @@ void main()
    vec4 nightSkyColor = texture(nightSky, -v3Direction);
    nightSkyColor = mix(nightColor, nightSkyColor, useCubemap);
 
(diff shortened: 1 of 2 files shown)


Azaezel:
May or may not want to follow up with shifting the colorspace of the result:
now
tosrgb
Figure leave that up to ya'll to debate till end of week. If no resolution comes of it, will just leave it alone.

rextimmy:
Don't clamp the output!

Azaezel:
> Don't clamp the output!

yeah. this'll kill off a couple clamps. only question left is srgb or no

rextimmy:
Keep it all in linear space like the rest of the pipeline, it will either be writing to a floating point (hdr) target or a srgb(ldr) one, so no need to do any manual conversions.

Azaezel:
yeah, way I lean as well since it's the least disruptive of existing projects, but did note the possibility when tinkering today so figured let folks weigh on in since this'll go in end of week.

PR #231, opened 2020-06-28
set GamepadButtonsGui.initMenuButtons(); call to self contained

Azaezel:


Azaezel:
superseded by the gui rewrite commits

PR #236, opened 2020-07-03
client cleanups

Azaezel:
part 1) order of operations on disconnectedCleanup


Azaezel:
should note as a followup, using https://github.com/Torque3DResources/FPSGameplay as a gameplay module, we do still seem to be getting 
unknown
which would seem to indicate there's still a few bits to adress either on the general core side, or something slipped in to that module overriding core cleanups

PR #237, opened 2020-07-03
safety check

Azaezel:
ensures
`   Knot *p1 = getKnot((S32)mFloor(t));
   Knot *p2 = next(p1);`
can't trip bad memory


lukaspj (approved):
Looks good to me

PR #240, opened 2020-07-07
fixed cpu detection on 64bit windows

JeffProgrammer:
(and removing nasty assembly for… win32).

Note that this still works for 32bit windows.


JeffProgrammer:
On the brightside, the SSE code can finally work now on 64bit :D (there is actually some used)

PR #251, opened 2020-07-15
Added slider to consoleDlg for bg alpha (Needs refinement)

zfbx:
So this works.. but looks kinda weird if you open the console before you open the editor. The GuiSliderCtrl should probably be copied (or better yet, moved) to the core directory so it can be used everywhere


Azaezel:
reminder since it's been a bit: last we talked, this needed that `GuiToolTipProfile` swapped to a generic profile so folks could ship the console without having to ship the whole tools suite

Azaezel:
to correct that, the profile in questions the `guiSliderProfile` one, not the tooltip. sorry bout the confusion

zfbx:
Okay, works without problems on my side now.

PR #259, opened 2020-07-26
Captures secondary window close events

Areloch:
Captures secondary window close events so secondary windows can be closed via hitting the x on the window itself.


Azaezel:
verified

PR #278, opened 2020-08-08
Suffix case adjustment for auto-import

Bloodknight:
Stops the engine from creating names with arbitrarily all caps suffixes to .taml and .cs files on importing assets via asset browser.


Azaezel:
resolved with #279

PR #317, opened 2020-09-22
Implements hook-look-up logic for shape assets to ShapeBaseData including autoimport handling

Areloch:
…ding autoimport handling

Also fixes issues with editing/spawning DBs via the asset browser


Azaezel:
pending time to beat this one up well, but a quick visual read over would suggest we toss in a writeflag/readflag   
shapeName = stream->readSTString();

   char buffer[256];
   stream->readString(buffer);
so we're only network transmitting the asset if that's valid

PR #323, opened 2020-09-30
shift pbrconfig to ORM in keeping with the prepoderance of industry standards

Azaezel:
having spoken with several endusers, artists and reviewing toolchains and augments to the systems in place, we'll be shifting the smooth|ao|metalness pbrconfigmaps on over to ao|roughness|metalness maps
first up, proper brdfTexture


Azaezel:
also includes math review from @rextimmy

PR #325, opened 2020-10-02
Fix cmake build on linux

lukaspj:
SDL 2.0.10 includes a bug that prevents SDL being built when the calling project is built in-tree, due to its' use of the global variables as opposed to the local variables when checking its requirement to be built out of tree.

https://github.com/GarageGames/Torque3D/pull/2383


lukaspj:
I just ported this over from the GarageGames repo, I don't have any way to verify that it works as I don't have Linux machine running atm, and won't for the foreseeable future. Would be great if someone could verify this on a Linux machine

Escalion:
I wrote the original PR and can confirm that it was tested and working.

Azaezel:
sorry. for clarity, that got fixed

PR #327, opened 2020-10-02
Fix global variables not being able to be used inside of a foreach$ loop

lukaspj:
This has been a bug since the old GG interpreter. Probably since foreach$ was introduced. You can now use foreach$ with globals.

https://github.com/GarageGames/Torque3D/pull/2361


lukaspj:
I can confirm that the issue exists, and this PR fixes it.

PR #330, opened 2020-10-02
Add test/debug skybox

lukaspj:
https://github.com/GarageGames/Torque3D/pull/2175


lukaspj:
Tested, works fine I guess. Should the one be laying down like that?
And is this needed?
image

JeffProgrammer (approved):

PR #333, opened 2020-10-03
Feature/badbehaviour

lukaspj:
Add bad behaviour system

_Note: I tried updating this to the module system but I haven't properly tested this, this should be seen as a PoC_'

Original PR from the GarageGames repo:
https://github.com/GarageGames/Torque3D/pull/2120


lukaspj:
Seems to work, but needs UI updates to play nicely with dark-theme and could use proper testing.

ghost:
BadBehavior itself works fine and is easy to integrate (just copy the folder), I'm using it for years, the only problem is, that the UI does not work, so I created my behavior trees using my code editor instead of the UI inside the engine, this is the reason I hesitated to make a pull request of it into the main repo.

The issue with BadBehavior see here: https://github.com/BadBehavior/BadBehavior_T3D/issues/2

I would just add it to the engine and have others fix the problem, I mean it is not the first unusable feature that was added to the engine without ever testing it.

PR #335, opened 2020-10-03
allows different waypoints

lukaspj:
I wanted to share this for some time, not sure if this is the best way to do it, but before it falls into the void :) this will allows to have at the same time different waypoint/markers, add more "WayPointMarker"....

Well you could always add from script, but this PR allows to add it from the World Editor.

See an example, at the right the green old shape. Now you can have different shapes and colors used for enemies, allies, objects.... I can do other PR with the markers I use, it have different shapes and colors, if required.
29991497-96c010aa-8f4d-11e7-9d28-f5b5f08c36a6
29991651-8de9e282-8f50-11e7-91a2-0d2836df54c7

Original PR from the GarageGames repo:
https://github.com/GarageGames/Torque3D/pull/2076

### DIFF
--- a/Templates/BaseGame/game/tools/worldEditor/gui/objectBuilderGui.ed.gui
+++ b/Templates/BaseGame/game/tools/worldEditor/gui/objectBuilderGui.ed.gui
@@ -1122,20 +1122,21 @@ function StaticShapeData::create(%data)
 
 function MissionMarkerData::create(%block)
 {
-   switch$(%block)
+   if(%block $= "SpawnSphereMarker")
    {
-      case "WayPointMarker":
-         %obj = new WayPoint() {
-            dataBlock = %block;
-            parentGroup = EWCreatorWindow.objectGroup;
-         };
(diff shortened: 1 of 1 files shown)


lukaspj:
The changes seems fairly straight-forward. I don't have time to properly test this, but can't seem the harm in it.

JeffProgrammer:
The changes seem to not be doing anything. It looks like equivalent code...also if you wanted to add more, you'd have to change the conditions now instead of just adding a new switch statement. Does this actually fix a bug?

lukaspj:
This is a QoL not a bug fix.

It seemingly insignificant, but right now it's a switch that defaults to a SpawnSphere unless it's set to one of two specific datablocks.
With this change, you can have different Waypoint Datablocks without hitting the default of spawnsphere

lukaspj:
Actually this could also be accomplished with just changing the switch statement afaict.
The idea is to allow arbitrary waypoint datablocks anyways

PR #339, opened 2020-10-03
ForceFOV is not working on GuiTSControl.

lukaspj:
The definition of frustum is already made. Changing mLastCameraQuery.fov on _internalRender is ignored. It has to be changed before frustum is defined.

Note: With this PR ForceFOV will not work on Render Stereo modes. I don't know it this is important...

https://github.com/GarageGames/Torque3D/pull/2018


lukaspj:
Looking at the old PR, it seems there might be some issue with stereo rendering. @Areloch does this ring any bells?

Azaezel:
yeah, main concern there's whether or not it'd break vr support

OTHGMars:
I verified this PR does fix the forceFOV field in GuiTSCtrl when rendering to a desktop window and does not alter the FOV when rendering to a HMD. I believe this is correct behavior. The FOV rendered in the HMD should always match the FOV reported by the device.

PR #341, opened 2020-10-03
breakShape() - remove parts created with invalid object box . 

lukaspj:
This fixes an invalid object box fatal assert that manifests when calling the Shapebase::blowUp() function on some (all?) .dts models. It appears this function is the only place in the engine that currently calls TsPartinstance::breakShape().

The issue might be model-specific. With some models, the routine will consistently generate a number of parts that produce an invalid object box when updateBounds() is called. This appears to happen when dealing with a denormalized min/max dimension so it is unclear if these parts are an artifact similar to empty parts. There might be an underlying issue in updateBounds() that could be cleaned up. Removing the invalid objects from the partList eliminates the crash while producing a satisfactory quantity of debris (when applied to a complex model).

This patch solves the crash for .DTS models, but the routine does not appear to produce any parts for .DAE models, and I have not looked in to this.

https://github.com/GarageGames/Torque3D/pull/1794

### DIFF
--- a/Engine/source/ts/tsPartInstance.cpp
+++ b/Engine/source/ts/tsPartInstance.cpp
@@ -107,10 +107,16 @@ void TSPartInstance::breakShape(TSShapeInstance * shape, S32 subShape, Vector<TS
    // update bounds (and get rid of empty parts)
    for (S32 i=0; i<partList.size(); i++)
    {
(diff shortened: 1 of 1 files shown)


lukaspj:
I fixed the code-styling issues that were mentioned in the original PR. Should be good to merge

PR #342, opened 2020-10-03
Replace uses of dStrCmp with new String::compare

lukaspj:
Moves from using dStrCmp to the new String::compare static functions keeps things cleaner, consistent, and works with intellisense.

This is the part of the move to replace usage of the old dStr___ functions for the more standard String namespace approach.

Original PR at GarageGames repo:
https://github.com/GarageGames/Torque3D/pull/1323


Azaezel:
10-1 this'll warrant a followup for any new ones

lukaspj:
@Areloch this one originates from you.
I have updated it to handle new cases in the 4.0 branch

PR #344, opened 2020-10-03
Added support to read PSD files.

lukaspj:
https://github.com/GarageGames/Torque3D/pull/975

I couldn't find the PostFXManager GUI in 4.0, so i removed those changes from this PR.

### DIFF
--- a/Engine/source/gfx/bitmap/loaders/bitmapPsd.cpp
+++ b/Engine/source/gfx/bitmap/loaders/bitmapPsd.cpp
@@ -1,0 +1,275 @@
+//-----------------------------------------------------------------------------
+// Copyright (c) 2012 GarageGames, LLC
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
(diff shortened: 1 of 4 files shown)


lukaspj:
Anyone with a PSD file that could verify this would be great :)

zfbx:
Either Github is being weird or i'm being dumb but it's not letting me submit a change suggestion on the file itself but https://github.com/lukaspj/Torque3D/blob/11d71aceaef6a3501d4d1e316d05dc8fdb933524/Engine/source/gfx/bitmap/loaders/bitmapPsd.cpp#L273 missing a `;` there at the end causing build to fail

zfbx:
Also I got it to work on the splash screen and guiBitmap controls without much problem but doesn't seem to wanna work as window icon. I tested 3 different difficulties of image:
[Red gradient to transparency](https://cdn.discordapp.com/attachments/609257102900199424/769488397793492993/gradient-red-transparent.psd)
[Overlapping shapes on different layers](https://cdn.discordapp.com/attachments/609257102900199424/769488398515044352/overlapping-shapes-transbg.psd)
[Overlapping shapes again but with text using a custom font and stroke effect](https://cdn.discordapp.com/attachments/609257102900199424/769488394518134794/overlapping-shapes-transbg-fancy-text-stroke.psd)

^I'm curious what happens if anyone else uses the font included psd

[Video](https://cdn.discordapp.com/attachments/416629987936829451/769489727954092052/zcYcm8l4Bi.mp4)

marauder2k7:
i know this is an older PR but PSD support was added with the STB PR, it cant handle separate layers though

PR #369, opened 2020-10-24
Added date/timestamp option to console log

zfbx:
Added support for prepending a fairly popular timestamp style Example: `[24/10/2020 05:38:39]Console message`
to use: `Con::useRealTimestamp = true;` 
Open to different name suggestions for the variable if it's not liked.
Also if it's requested could make date and/or time options but I see that getting to be a little messy in code. Also plays nice with `Con::useTimestamp`

### DIFF
--- a/Engine/source/console/console.cpp
+++ b/Engine/source/console/console.cpp
@@ -621,7 +621,7 @@ static void _printf(ConsoleLogEntry::Level level, ConsoleLogEntry::Type type, co
    {
       Platform::LocalTime lt;
       Platform::getLocalTime(lt);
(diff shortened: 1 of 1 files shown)


dottools:
I would prefer the universal `YYYY-MM-DD hh:mm:ss` (ISO 8601) format over the region specific `MM/DD/YYYY` -or- `DD/MM/YYYY` date format. Also needs a space ` ` between closing bracket `]` and message.

zfbx:
Yeah, I was really tired last night and thought it was universal date order.. oops. also I went against the space at the end at first because the other timestamp doesn't use a space and instead of adding the space to the end of that timestamp I put a check if either timestamp is used to add one so if both timestamps are used they'll be like `[2020-10-24 10:00:00][+   3.750] Message`

PR #375, opened 2020-11-03
Adds handling to the path manager so it can deal with both looping and non-looping paths

Areloch:
Adds handling to the path manager so it can deal with both looping and non-looping paths
As provided by Ubiq from their Action Adventure Kit

### DIFF
--- a/Engine/source/scene/pathManager.cpp
+++ b/Engine/source/scene/pathManager.cpp
@@ -102,6 +102,7 @@ void PathManagerEvent::pack(NetConnection*, BitStream* stream)
    stream->write(modifiedPath);
    stream->writeFlag(clearPaths);
    stream->write(path.totalTime);
+   stream->write(path.looping);
    stream->write(path.positions.size());
 
 
@@ -131,6 +132,7 @@ void PathManagerEvent::unpack(NetConnection*, BitStream* stream)
    stream->read(&modifiedPath);
    clearPaths = stream->readFlag();
    stream->read(&path.totalTime);
+   stream->read(&path.looping);
 
    U32 numPoints;
    stream->read(&numPoints);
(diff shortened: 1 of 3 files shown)


Engine/source/scene/pathManager.cpp
```diff
@@ -299,15 +303,33 @@ void PathManager::getPathPosition(const U32 id,
 
    // Ok, query holds our path information...
    F64 ms = msPosition;
-   if (ms > mPaths[id]->totalTime)
-      ms = mPaths[id]->totalTime;
+
+   //Looping vs. non-looping splines
+   F32 msTotal;
+   if (mPaths[id]->looping)
+      msTotal = mPaths[id]->totalTime;
```
lukaspj, line 310: I know that this is similar to how the code was written before, but I believe our official code-style is to always include braces?

Engine/source/scene/pathManager.cpp
```diff
@@ -299,15 +303,33 @@ void PathManager::getPathPosition(const U32 id,
 
    // Ok, query holds our path information...
    F64 ms = msPosition;
-   if (ms > mPaths[id]->totalTime)
-      ms = mPaths[id]->totalTime;
+
+   //Looping vs. non-looping splines
```
lukaspj, line 307: Nit-picking here but I prefer having a space between // and comment text:

```suggestion
   // Looping vs. non-looping splines
```

Engine/source/scene/pathManager.cpp
```diff
@@ -299,15 +303,33 @@ void PathManager::getPathPosition(const U32 id,
...
+   F32 msTotal;
```
lukaspj, line 308: Shouldn't this be a `U32` instead?

```suggestion
   U32 msTotal;
```

Engine/source/scene/pathManager.cpp
```diff
@@ -430,5 +466,83 @@ bool PathManager::readState(BitStream* stream)
...
+   //2) Otherwise, divide the spline-segment into S sub-segments (S+1 points to test)
+   //2) Test these points against the given point and find the closest of them
+   //4) Recurse within the 2 segments surrounding the closest point
 
+   //NOTE: In a case where the spline comes near the point multiple times, the wrong local 
+   //minima of the spline may be chosen early on and then refined, like polishing a turd :)
+
+   PROFILE_START(PathManager_getClosestTimeToPoint);
+
+   F64 totalTime;
```
lukaspj, line 485:
```suggestion
   U64 totalTime;
```

Azaezel:
went ahead and merged this because it looks like it contains a dependency for https://github.com/Torque3DResources/ActionAdventureKit and didn't display any particular nastiness for the testbed I'm using at least despite the concerns, though may indeed merit followup

PR #380, opened 2020-11-19
TAML Serialization Improvements

lukaspj:
Just so it isn't forgotten, this is some necessary improvements, in order to properly support serializing complex objects with TAML.
More examples here:
https://github.com/GarageGames/Torque3D/pull/1221

NB: This does not currently build due to a type issue with some strings or something.


Azaezel:
should note the original pull on this one references a partial incorporation. might be some interference there?

lukaspj:
I'm not sure what you mean @Azaezel can you link to the comment about that?

PR #389, opened 2020-12-01
variation on #387 that also introduces errorcodes

Azaezel:


Azaezel:
will do up a clean once this is better sorted

PR #402, opened 2020-12-09
during preload, autoreplace direct file refs with assetIDs

Azaezel:
riffing off of updateMaterialsScript, during shapebasedata preload, spools up a PersistenceManager instance, if shapeAssetId is unset, sets the object dirty, and saves it off once processing is completed


Azaezel:
https://github.com/TorqueGameEngines/Torque3D/blob/d50f1f0b811b92f7ec51c3c7b5912a6a6657355a/Engine/source/console/persistenceManager.cpp#L1021 seems to be causing issues with resaving on the fly. example: https://github.com/Torque3DResources/FPSGameplay/blob/master/datablocks/weapons/Lurker.cs#L314

PR #405, opened 2020-12-12
WIP: C# CMake module

lukaspj:
Depends on https://github.com/TorqueGameEngines/Torque3D/pull/406

Allows easy setup of a C#-based game project.

Read more about T3DSharp in the repo's wiki:
https://github.com/lukaspj/T3DSharp/wiki

Steps to test:

**Windows:**
_(Currently this assumes Visual Studio or at the very least MSBuild being used)_
 1. Git Clone this branch
 1. Git Clone https://github.com/lukaspj/T3DSharp-CoinCollection into `game/data/CoinCollection`
 1. CMake configure
 1. Check `TORQUE_CSHARP` and `USE_SOLUTION_FOLDERS` (recommended)
 1. CMake Generate
 1. Open Project in Visual Studio and build

**[WIP] UNIX**:
 1. [Install Dotnet](https://docs.microsoft.com/en-us/dotnet/core/install/linux-scripted-manual)
 1. Git Clone this branch
 1. Git Clone https://github.com/lukaspj/T3DSharp-CoinCollection into `game/data/CoinCollection`
 1. Run CMake with `TORQUE_CSHARP=ON` and `TORQUE_CXX_FLAGS_LIBS="-fPIC"`
    1. `cmake ../../../.. -DTORQUE_APP_NAME=LinuxTest -DCMAKE_BUILD_TYPE=Release -DTORQUE_CSHARP=ON -DTORQUE_CXX_FLAGS_LIBS="-fPIC"`
    2. `cmake --build . --parallel 8` (set parallelization here for quicker builds)
 1. Run `LD_LIBRARY_PATH=. ./T3DSharpGame` in the `game` folder

PR #406, opened 2020-12-12
Parametrize script extension, default to 'tscript'

lukaspj:
Defaults to `.tscript` and that is how it's stated in the comments around the code now.
But if one prefers `.cs`, they can fairly easily change it back.


Azaezel:
Slated up as the next major overhaul step as prepwork for #405, 

Observations thus far include a desire to tie the cmake entry of TORQUE_SCRIPT_EXTENSION to the root script "${projectOutDir}/main.cs(.tscript)", likely by pulling concatenation tricks https://github.com/TorqueGameEngines/Torque3D/blob/612a3e3358423131401c125725d7e9bef342065b/Tools/CMake/torque3d.cmake#L644-L645 of the forms described https://stackoverflow.com/questions/18001198/how-do-you-concatenate-string-in-cmake to get the resulting filename to write to.

We'll likely also want to cook up crossplat cli batch files so pre-existing projects can mass convert their existing scripts,

And it'd likely be a good idea to have a prefab copy of the project manager lurking around to throw this at (and might even be able to solve point 2 by leveraging a shiny button in point 3 depending on how far along that is)

lukaspj:
@Azaezel I updated the CMake thingy.

Also, made the core-template agnostic towards script extensions through the `$TorqueScriptFileExtension` global.
Finally I modified `execFile` so if you say `exec("scripts/foo");` It will automatically try appending an extension.

I'm not sure if a conversion tool belongs in this PR nor am I sure I'm in support of one really, since the string "cs" can be ambiguous and you might encounter false positives. 
But if one were to be made I'd suggest something along the lines of (pseudocode):
```bash
Find All ".cs" in files (*.cs;*.mis;*.gui):
  Replace with ".\" @ $TorqueScriptFileExtension @ \""

Find All "@ \"\"" in files (*.cs;*.mis;*.gui):
  Replace with ""
```

But that will probably not catch all the things.

Another solution would be to simply ask people to set their preferred extension, and if they don't want to use `.tscript` they can just keep using `.cs` for legacy projects.

Azaezel:
overall checks out though I should note
https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Templates/BaseGame/game/core/Core.module#L5
needs to read
 ScriptFile="Core"
not
 ScriptFile="./Core"
to function

lukaspj:
@Azaezel I fixed the Core module, sorry for that one

PR #410, opened 2020-12-14
pad isFile search string buffer

Azaezel:
we were hitting a few cases where the isfile comparison internal buffer was retrieving the raw file location when testing for the relative and going out of bounds. this ensures there's enough preallocated space


Azaezel:
the bug that lead to this being  needed has been resolved, so this one's pretty much a paranoia step. up to yall whether to toss it in, or kill it.

PR #423, opened 2020-12-26
Fix ORM maps in terrain textures

### LINKED ISSUE #422: Terrain Texture: ORM map not applied properly
There is some bad math in the calculation of the ORM map, only the first layer with ORM is applied properly:
image

```hlsl
// Composite Matinfo map 0
float3 matinfoCol0 = lerp(float3(1.0,1.0,0.0),ormMapArray.Sample(ormMapSampler, float3(IN.detCoord0.xy,detailIdStrengthParallax[0].x)).rgb, detailBlend0);
OUT.col2 = float4(0.0,matinfoCol0);

// Composite Matinfo map 1
float3 matinfoCol1 = ormMapArray.Sample(ormMapSampler, float3(IN.detCoord1.xy, detailIdStrengthParallax[1].x)).rgb*detailBlend1;
OUT.col2.gba += matinfoCol1;
```

> Wouldn't that set it to (1,1,0) after the first composite map, and then when you add the second composite map, those would get above 1 since you add it?

lukaspj:
This fixes some bad maths in terrain texture's ORM map.

Worth noting that edges between textures still seem pretty pronounced, but these fixes at least makes it somewhat passable.

image

Fixes #422

### DIFF
--- a/Engine/source/terrain/glsl/terrFeatureGLSL.cpp
+++ b/Engine/source/terrain/glsl/terrFeatureGLSL.cpp
@@ -1260,7 +1260,7 @@ void TerrainORMMapFeatGLSL::processPix(Vector<ShaderComponent*> &componentList,
 
 	// search for material var
 	Var * ormConfig;
-	OutputTarget targ = RenderTarget1;
+	OutputTarget targ = DefaultTarget;
 	if (fd.features[MFT_isDeferred])
 	{
 		targ = RenderTarget2;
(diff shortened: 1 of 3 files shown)


lukaspj:
I fixed it:
image


Should be ready now.

DirectX works splendidly, OpenGL still has some issues, but I'm not convinced it's related to these changes. OpenGL in general is not really working very well.

Azaezel:
this'll be covered by https://github.com/TorqueGameEngines/Torque3D/pull/424

PR #427, opened 2020-12-29
leverage new SDL_OpenURL( adress ); functionality

Azaezel:


Azaezel:
https://cdn.discordapp.com/attachments/416629987936829451/791112609683865610/mainMenu.gui test file

JeffProgrammer:
Can we make platform:: call just call the SDL one internally? That way if TORQUE_SQL is disabled, it don't go pearshaped in multiple places?

Azaezel:
Well, fars it goes it already does, and we've been actively trying to kill off platform code that sdl already handles whenever we get freetime/a bug report leads us to finding code best handled that way... If anything, my instinct is to say remove the TORQUE_SQL flag at this point, and keep it simple.

rextimmy:
SDL shouldn't be exposed to the rest of the engine like this. As mentioned by jeff it should be internally implemented with Platform::openWebBrowser. Also, generally removing platform code and replacing with raw sdl code is not a good idea at all, nightmare if anyone wants to bring t3d to a platform that has either poor sdl or no sdl support at all, take a look at the consoles for example.

Azaezel:
fair points. @Areloch, up to you, #431 prepped if you want to maintain subdivisions as they exist now. could also see throwing the functionality in https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Engine/source/platform/platform.h#L347 and just letting folks override if they come to that for unsupported platforms to cut down on redundancies.

rextimmy:
just add a new file called sdlPlatform.cpp in the paltformSDL folder and add him in there, I'm sure over time there can be more 'sdl replacements' added in there as well.

Azaezel:
superseded by  #432

PR #430, opened 2020-12-31
mac fixes

Azaezel:
1) add sdl token for referencing the parent directory of the aplication bundle when determining the root dir
2) shape and mateiral asset macro corrections for a few stray ## redundencies (space, ->, . ect are all conidered token sperators in and of themselves)


JeffProgrammer:
works.

PR #437, opened 2021-01-04
debate point: should we allow terrain blending to have layers recede below 0

Azaezel:
Can see several arguments both for and against this proposed follow-up to #424. Vid of the effect https://cdn.discordapp.com/attachments/580242040730943509/795615167630802974/neoTerrain_2021-01-04_05-29-10-799.mp4

up to yall.


rextimmy:
There is a cool new feature with github called discussions, saves creating issues or PRs so you can well, discuss stuff :) https://docs.github.com/en/free-pro-team@latest/discussions/quickstart#creating-a-new-discussion

PR #440, opened 2021-01-04
corrects handling of multiple modlues with the same filename

Azaezel:
the only time we want to update in place is when an existing match is present, and the incomming file ref is exclusive


Azaezel:
#446 covers this too

PR #448, opened 2021-01-11
general spawnspheres

Azaezel:
add a %this.registerMissionObject( "SpawnSphere", "General Spawn Sphere", "GeneralDropPoint" ); to the level creator folder
this one allows you to plug in an object-instance type as well as any datablock, not just speified subtype and run the %obj.spawnObject(""); command on it to execute generic object creation based on the config of that SpawnSphere


Azaezel:
steps to test:
 go to the level folder, 
 plug in a class that takes a datablock, and a valid datablock
 use the command objectIDnum..spawnObject(""); in the console to verify it spawns the thing

PR #435, opened 2021-01-03
Loading from zipped game directories.

OTHGMars:
This PR addresses the issues mentioned in #22.
Also includes fixes for Ogg Vorbis, NavMesh and recursive searching into mounted sub-directories.

PR #449, opened 2021-01-11
Loading from zipped game directories.

OTHGMars:
This PR is a replacement for PR #435. This is a cleaner fix for the zipped sound files and solves the problem with concurrent zip file access that was exposed when reading forests.


Azaezel:
Just to keep folks in the loop: this works as advertised, but there's a slight bug with assetification we need to roll in to compensate where it's generating duplicates if multiple assets share a base file name from which to generate.

Beyond that, will note the suggestion of using a cmake entry to set https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Engine/source/core/util/zip/zipArchive.h#L40, likely in the usual spot https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Tools/CMake/torqueConfig.h.in but that can be done as followup.

PR #450, opened 2021-01-11
provides a general Playgui_onWake callback for modules

Azaezel:
any module with a modulename::Playgui_onWake now has the capacity to layer thier gui elements on playgui


Azaezel:
test procedure:
create or augment an existing module such as ExampleModule with a gui element
create within that module a file, or augment an existing one with for instance:

function ExampleModule::Playgui_onWake()
{
    Canvas.pushDialog(someGuiElement);
}

PR #456, opened 2021-01-22
Fix for transparency in splash images

zfbx:
This comes from a conversation back in October 2020 source from this section of the discord#coding channel https://discord.com/channels/358091480004558848/416629987936829451/765339207878836224 

This is the simplified version that just works automagically based on the image, there was more work done using variables and such from marauder2k9 here https://discord.com/channels/358091480004558848/416629987936829451/765345319478624306 that I just thought was a little too much extra for what i personally think should be just a simple "it works". 

**This has NOT been tested AFAIK on Mac or Linux** *I also don't know if it has to be or not, idk how much voodoo sdl does


Azaezel:
will need to encapsulate portions of this one in an #ifdef TORQUE_SDL block

PR #457, opened 2021-01-24
augment datablock file handling to include references with no suffixes

Azaezel:


Azaezel:
verification method: in a module like https://github.com/Torque3DResources/FPSGameplay/blob/master/FPSGameplay.cs#L73 , try 
%this.registerDatablock("./datablocks/player"); instead so it doesn't care what the extension used is

PR #466, opened 2021-02-08
set prefab and makemesh origins to the biggest model

Azaezel:
allows one to preserve snap points


Azaezel:
test procedure: with multiple objects selected, and "object center" instead of "bound center" selected, try and bake a mesh/make a prefab. should see the result as the origin of your largest object
pics for clarity:
https://cdn.discordapp.com/attachments/580242040730943509/808479698749358130/unknown.png
vs
https://cdn.discordapp.com/attachments/580242040730943509/808479741148528660/unknown.png

PR #474, opened 2021-02-24
Fixes display of internal names on objects shown in guiTreeViewCtrl

Areloch:
Test case:
Add internal name to an object that can be viewed in a guiTreeViewCtrl and then ensure that the display shows as the proper objectname[internalName]

### DIFF
--- a/Engine/source/gui/controls/guiTreeViewCtrl.cpp
+++ b/Engine/source/gui/controls/guiTreeViewCtrl.cpp
@@ -560,7 +560,7 @@ void GuiTreeViewCtrl::Item::getDisplayText(U32 bufLen, char *buf)
                if (mState.test(Item::Marked))
                   dSprintf(ptr, len, " *[%s]", pInternalName);
                else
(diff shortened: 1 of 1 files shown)


Engine/source/gui/controls/guiTreeViewCtrl.cpp
```diff
@@ -560,7 +560,7 @@ void GuiTreeViewCtrl::Item::getDisplayText(U32 bufLen, char *buf)
                if (mState.test(Item::Marked))
                   dSprintf(ptr, len, " *[%s]", pInternalName);
                else
-                  dSprintf(ptr, len, " [%s]", pObjName);
+                  dSprintf(ptr, len, " [%s]", pInternalName);
```
dottools, line 563: I think this way is a bit cleaner:
```suggestion
               dSprintf(ptr, len, " %s[%s]", (mState.test(Item::Marked))? "*":"", pInternalName);
```

PR #478, opened 2021-03-18
Fixes a resolution switching issue when the game uses **only** OpenGL…

Bloodknight:
… as the renderer

Canvas.tscript fixes provided by @OTHG_Mars

### DIFF
--- a/Engine/source/gfx/gl/sdl/gfxGLDevice.sdl.cpp
+++ b/Engine/source/gfx/gl/sdl/gfxGLDevice.sdl.cpp
@@ -58,7 +58,16 @@ void EnumerateVideoModes(Vector<GFXVideoMode>& outModes)
       GFXVideoMode outMode;
       outMode.resolution.set( mode.w, mode.h );
       outMode.refreshRate = mode.refresh_rate;
-      outMode.bitDepth = SDL_BYTESPERPIXEL( mode.format );
+
+      // BBP = 32 for some reason the engine knows it should be 32, but then we
+      // add some extra code to break what the engine knows.
+      //outMode.bitDepth = SDL_BYTESPERPIXEL( mode.format );     // sets bitdepths to 4
+      //outMode.bitDepth = SDL_BITSPERPIXEL(mode.format);        // sets bitdepth to 24
+      
+      // hardcoded magic numbers ftw
+      // This value is hardcoded in DX, probably to avoid the shenanigans going on here
+      outMode.bitDepth = 32;                                   
+
       outMode.wideScreen = (mode.w / mode.h) > (4 / 3);
       outMode.fullScreen = true;
       
--- a/Templates/BaseGame/game/core/gui/scripts/canvas.tscript
+++ b/Templates/BaseGame/game/core/gui/scripts/canvas.tscript
@@ -155,6 +155,10 @@ function GuiCanvas::prefsToModeStr(%this)
 
 function GuiCanvas::checkCanvasRes(%this, %mode, %deviceId, %deviceMode, %startup)
 {
+   // Toggle for selecting the borderless window allowed sizes. Set true to allow
+   // boderless windows to be less than the device res.
+   %allowSmallBorderless = true;
+
    %resX = getWord(%mode, $WORD::RES_X);
    %resY = getWord(%mode, $WORD::RES_Y);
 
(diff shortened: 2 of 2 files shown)


Engine/source/gfx/gl/sdl/gfxGLDevice.sdl.cpp
```diff
@@ -58,7 +58,16 @@ void EnumerateVideoModes(Vector<GFXVideoMode>& outModes)
...
-      outMode.bitDepth = SDL_BYTESPERPIXEL( mode.format );
+
+      // BBP = 32 for some reason the engine knows it should be 32, but then we
+      // add some extra code to break what the engine knows.
+      //outMode.bitDepth = SDL_BYTESPERPIXEL( mode.format );     // sets bitdepths to 4
+      //outMode.bitDepth = SDL_BITSPERPIXEL(mode.format);        // sets bitdepth to 24
+      
+      // hardcoded magic numbers ftw
+      // This value is hardcoded in DX, probably to avoid the shenanigans going on here
+      outMode.bitDepth = 32;
```
lukaspj, line 69: I'd prefer if we figured out what's going wrong :)

Templates/BaseGame/game/core/gui/scripts/canvas.tscript
```diff
@@ -155,6 +155,10 @@ function GuiCanvas::prefsToModeStr(%this)
 
 function GuiCanvas::checkCanvasRes(%this, %mode, %deviceId, %deviceMode, %startup)
 {
+   // Toggle for selecting the borderless window allowed sizes. Set true to allow
+   // boderless windows to be less than the device res.
```
lukaspj:
```suggestion
   // borderless windows to be less than the device res.
```

Templates/BaseGame/game/core/gui/scripts/canvas.tscript
```diff
@@ -202,27 +209,34 @@ function GuiCanvas::checkCanvasRes(%this, %mode, %deviceId, %deviceMode, %startu
    return false;
 }
 
-// Find the best video mode setting for the device and display mode
+// Find the best video mode setting for the device and display mode.
+// "Best" is the largest resolution that will fit at highest refresh rate.
```
lukaspj, line 213: Thanks for improving the comments! :)

I believe TorqueScript supports doxygen-ish comments if you do three slashes, but it's not really a heavily-used feature

Templates/BaseGame/game/core/gui/scripts/canvas.tscript
```diff
@@ -202,27 +209,34 @@ function GuiCanvas::checkCanvasRes(%this, %mode, %deviceId, %deviceMode, %startu
...
+      if ((%resX > %deviceRect.x) || (%resY > %deviceRect.y) ||
+         (%resX < $Video::minimumXResolution) || (%resY < $Video::minimumYResolution))
          continue;
 
-      return %testRes;
+      if (((%bestRes $= "") || (%resX > getWord(%bestRes, $WORD::RES_X)) ||
+            (%resY > getWord(%bestRes, $WORD::RES_Y))) ||
+         ((%resX == getWord(%bestRes, $WORD::RES_X)) && (%resY == getWord(%bestRes, $WORD::RES_Y)) &&
+            (%rate > getWord(%bestRes, $WORD::REFRESH))))
+         %bestRes = %testRes;
```
lukaspj, line 238: This is... This is fairly difficult to follow..

So:
If bestRes is not set OR resX is greater than the bestRes OR resY is greater than bestRes OR (resX is the same as bestRes AND resY is the same as bestRes AND rate is higher) then testRes = bestRes 

Would the following work?

```suggestion
      %totalRes = %resX * %resY;
      %totalBestRest = getWord(%bestRes, $WORD::RES_X) * getWord(%bestRes, $WORD::RES_Y);
      if (%totalRes < %totalBestRes) continue;
      if (%totalRes == %totalBestRes && %rate <= getWord(%bestRes, $WORD::REFRESH)) continue;
      %bestRes = %testRes;
```

Engine/source/gfx/gl/sdl/gfxGLDevice.sdl.cpp
JeffProgrammer: interesting that the note is that DirectX has it hardcoded as 32bits...I wonder why SDL is incorrectly detecting the bitdepth. We'd PROBABLY be okay with hardcoding it as...any computer since 2000 I think would support 32bit color depth, but it would be interesting to see why SDL is causing issues.

Azaezel: Fwiw, did try and do a trace on that at the time of submission, and ended up very much lost in the weeds. end of the day, it's a step towards consistent behavior across platforms a least

Templates/BaseGame/game/core/gui/scripts/canvas.tscript
Azaezel: Hesitant to add resolution checks in the form of %totalRes = %resX * %resY; in case that plays up for folks trying to add phone et al support down the line. we might not have plans to pursue it at time of writing, but lets not make headaches for future-us, eh?

Azaezel:
Barring strong objections, calling this one good enough. May merit a follow-up, but still the best options we got in a reasonable amount of time.

PR #479, opened 2021-03-24
Implement Unit Test Suite for TorqueScript.

JeffProgrammer:
Added unit tests for TorqueScript. This will give us more confidence as changes are made to the script interpreter.


lukaspj (approved):
Only improvement I can think of would be more tests, but that doesn't take away from these changes.
Greate work! :)

PR #483, opened 2021-04-09
Updates the masterserver domain referenced in the default scripts, pointing it to the new torque3d master server

Areloch:
Test: Affirm that querying the internet on the join menu polls the master server successfully.


lukaspj (approved):
Code looks fine

PR #484, opened 2021-04-12
form steve yorkshire: mat editor save extension fix

Azaezel:


Templates/BaseGame/game/tools/materialEditor/scripts/materialEditor.ed.tscript
```diff
@@ -2551,7 +2551,7 @@ function MaterialEditorGui::saveCompositeMap(%this)
     %saveAs = "";
     %dlg = new SaveFileDialog()
     {
-        Filters        = "png";
+        Filters        = "PNG File (*.png)|*.png|";
```
JeffProgrammer: is this the correct filter? It looks like its going to be changed to `"PNG File (*.png)|*.png|"`. It might be fine, but it looks weird.

PR #487, opened 2021-05-10
Add support for Apple Silicon

JeffProgrammer:
This adds support for running natively on the M1 Mac processors.


Engine/lib/convexDecomp/NvThreadConfig.cpp
```diff
@@ -135,6 +135,8 @@ void tc_spinloop()
       // Pause would do nothing on the Xbox. Threads are not scheduled.
    #elif defined( _WIN64 )
       YieldProcessor( );
+   #elif (defined( __arm64__ ) && defined( __APPLE__ )) || defined( __arch64__ )
+      asm( "yield" );
```
dottools: That does not look like a good idea. Need to use an OS specific yield function call in order to cause the scheduler to intentionally yield/swap out the current process/thread effectively.

JeffProgrammer: I'll replace it with pthread_yield()

PR #488, opened 2021-05-14
Window resolution options

OTHGMars:
Limits fullscreen windows to resolutions that match the monitor aspect ratio.
Adjusts borderless window default size to be the monitor usable area on Mac and Linux. This prevents the window from rendering behind the taskbar or titlebar on those platforms.


lukaspj:
Haven't got around to test it, but code wise it looks brilliant

PR #489, opened 2021-06-02
expose a zip file password cmake config option

Azaezel:


Engine/source/core/util/zip/zipArchive.h
```diff
@@ -235,7 +233,7 @@ class ZipTempStream;
 /// exists and may be released as a resource in the future.
 /// 
 /// To set the password used for zips, you need to modify the #DEFAULT_ZIP_PASSWORD
-/// define in core/zip/zipArchive.h. This password will be used for all zips that
+/// define in torqueConfig.h. This password will be used for all zips that
```
lukaspj, line 236: Well.. Shouldn't this say "you need to modify the configuration in CMake" or something?

lukaspj (approved):

Azaezel:
If you're reading it in the source, odds are ya already built it. The proposed cmake setup at present massages the config file included source side as opposed to burying it in a project config entry. Hence pointing at where it would be at that point in time.

PR #493, opened 2021-06-11
shadowmap validator tweaks

Azaezel:
set the minimum range to draw a shadow to 25 units, and the max to 10k (context, default draw distance is 1k)
'double' the max texture size allowed to 8192^2 (was 4096^2) to allow folks to throw more machine at it if they really *really* wish to


lukaspj (approved):
Looks good

PR #495, opened 2021-06-18
TorqueScript Interpreter 2.0

JeffProgrammer:
- Suite of acceptance tests in C++
- Function calls are up to 5x faster (with or without parameters)
- Method calls are up to 10x faster (with or without parameters)
- Arithmetic operations are up to 3x faster
- Parameter passing is much faster
- Less conversions between strings and numbers
- Possibly less memory usage as every internal variable is now 16 bytes (previously 32 bytes)
- Peephole optimizer (some constant folding and strength reduction)
- 99% backwards compatible syntax wise, most current scripts should just work (tm)\*
- Literally *hundreds* of bug fixes in the base template

\* Breaking Changes: 
- local variables are no longer allowed to be used outside of function
- nextToken() requires a global variable for the second parameter i.e. %str = nextToken(%str, $MyToken, ",");
- Anonymous function support has been removed, if someone has code in production that really needs this feature, let me know.
- eval() calls must return its value to be used, can't set a local variable in eval as local vars can't be used outside of a function
- Local variables must be assigned a value before they are looked up

Happy Testing and Enjoy, this was 3 months of hard work!


dottools:
Awesome!

lukaspj:
> Local variables must be assigned a value before they are looked up

What happens if you try to look it up before it is assigned a value?

JeffProgrammer:
> > Local variables must be assigned a value before they are looked up
> 
> What happens if you try to look it up before it is assigned a value?

You will get a compiler error when compiling the script. An Assert happens so you can fix it telling you the exact line it fails on. The compiler is smart enough to detect it.

PR #497, opened 2021-06-19
particle emitter bounds box fix

Azaezel:
take the distance particles can move over thier lifetime into account for emitter bounds.
in addition, don't zero out the y axis


lukaspj:
Looks good, only thought here would be how expensive is it to update the bounding box?
Could we store the min and max position here:
https://github.com/TorqueGameEngines/Torque3D/blob/cf0c7a2d1124cddc488fcfc5c70ebeb3f6d42391/Engine/source/T3D/fx/particleEmitter.cpp#L1825

And update the bounding box every tick? Maybe every 10th tick or something like that?

Azaezel:
will look into that and treat this as a fallback then

lukaspj (approved):
Edit: Typo

PR #498, opened 2021-06-22
Script extension assignment.

OTHGMars:
Sets extension on installed script files.
Enables .torsion file to be reconfigured after extension change.
Prevents copy of main.tscript.in.


Tools/CMake/torque3d.cmake
```diff
@@ -645,11 +645,13 @@ else()
   endif()
 endif()
 
-if(NOT EXISTS "${projectOutDir}/${PROJECT_NAME}.torsion")
-    CONFIGURE_FILE("${cmakeDir}/template.torsion.in" "${projectOutDir}/${PROJECT_NAME}.torsion")
-endif()
-if(EXISTS "${CMAKE_SOURCE_DIR}/Templates/${TORQUE_TEMPLATE}/game/main.${TORQUE_SCRIPT_EXTENSION}.in")
-    CONFIGURE_FILE("${CMAKE_SOURCE_DIR}/Templates/${TORQUE_TEMPLATE}/game/main.${TORQUE_SCRIPT_EXTENSION}.in" "${projectOutDir}/main.${TORQUE_SCRIPT_EXTENSION}")
+CONFIGURE_FILE("${cmakeDir}/template.torsion.in" "${projectOutDir}/${PROJECT_NAME}.torsion")
```
lukaspj: Fair warning, this would overwrite any changes the developer made to their Torsion project. This might not be desirable..
Could we regex the script extension in, instead? Or possibly have a "cmake.torsionconfig" file or something like that

Tools/CMake/torque3d.cmake
```diff
@@ -645,11 +645,13 @@ else()
...
+if(EXISTS "${CMAKE_SOURCE_DIR}/Templates/${TORQUE_TEMPLATE}/game/main.tscript.in")
+    CONFIGURE_FILE("${CMAKE_SOURCE_DIR}/Templates/${TORQUE_TEMPLATE}/game/main.tscript.in" "${projectOutDir}/main.${TORQUE_SCRIPT_EXTENSION}")
+else()
+    if(EXISTS "${CMAKE_SOURCE_DIR}/Templates/${TORQUE_TEMPLATE}/game/main.${TORQUE_SCRIPT_EXTENSION}.in")
```
lukaspj, line 654: I would prefer swapping these if-statements so main.cs.in has precedence over main.tscript.in if you are targeting main.cs
But I don't really see the use-case here tbh

Tools/CMake/torque3d.cmake
OTHGMars: The reason for removing the `if(Exists...` check is because when using cmake gui, the .torsion file gets created on the first configure pass after supplying a project name. This is before the user has had a chance to edit the script extension, so torsion is always setup for the .tscript extension regardless of the value of TORQUE_SCRIPT_EXTENSION. But, I agree, a better solution needs to be found for preventing user changes from getting overwritten.

Tools/CMake/torque3d.cmake
OTHGMars: The way it was originally written would only configure a main._scriptExension_ file if the template contained the script extension that was configured. Since the stock templates will all contain .tscript.in, I made that the first case and left the original as secondary. The only use case for the second is older templates that still contain main.cs.in.

lukaspj: Ah that makes sense! :)

Tools/CMake/torque3d.cmake
lukaspj: Hmm what about a warning?
I mean, sure probably no one reads the warnings, but it'd be nice with a little warning.

lukaspj (approved):

JeffProgrammer (approved):

PR #499, opened 2021-07-01
mac compilation and standarization fixes

Azaezel:
compresses void tc_spinloop() for both unix derivatives and apples to utilize pthread_yield_np();
mac only tweaks to Platform::getUserHomeDirectory() to return the ~/documents directory for parity with windows results

### DIFF
--- a/Engine/lib/convexDecomp/NvThreadConfig.cpp
+++ b/Engine/lib/convexDecomp/NvThreadConfig.cpp
@@ -129,17 +129,15 @@ void   tc_sleep(NxU32 ms)
 
 void tc_spinloop()
 {
-   #ifdef __linux__
-      asm ( "pause" );
-   #elif defined( _XBOX )
-      // Pause would do nothing on the Xbox. Threads are not scheduled.
-   #elif defined( _WIN64 )
-      YieldProcessor( );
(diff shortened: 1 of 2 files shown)


JeffProgrammer (approved):

Azaezel:
Interesting... we compressed what seemed like redundant flags since https://www.unix.com/man-page/osf1/3/pthread_yield_np/ turned up on a search of standard manual pages.

To be 100% sure, does replacing the block
```
#elif defined( __APPLE__ )||(__linux__)
   pthread_yield_np();
```
with
```
#elif defined( __APPLE__ )
   pthread_yield_np();
#elif defined(__linux__)
   pthread_yield();
```
cover your case?

SuperUserNameMan:
Yes it compiles.

According to official Linux kernel documentation here :  https://man7.org/linux/man-pages/man3/pthread_yield.3.html
- ` pthread_yield() ` is deprecated and non-standard, and should be replaced with ` sched_yield() `.

It compiled, but I could not test because there are currently some bugs into the default template tscripts ...

Azaezel:
Ok, running by the options with our other unix-derivative testers to ensure we're not causing havoc, hopefully have a concrete PR within the next couple days.

Issue #501, opened 2021-07-18
can't compile 4.0 on LinuxMint 20, because no `pthread_yield_np()`


SuperUserNameMan:
Hello,

Compilation of version dev 4.0 fails here :

` convexDecomp/NvThreadConfig.cpp:137:4: error: ‘pthread_yield_np’ was not declared in this scope; did you mean ‘pthread_yield’? `

System : LinuxMint 20 Ulyana, 64bits

edit : related to PR #499

SuperUserNameMan:
incoming PR here #507

Azaezel:
largely addressed, but leaving this ticket live in case of issues related to the commit.

JeffProgrammer:
are we good to close this now?

PR #507, opened 2021-07-19
adress #501 - thread oversight.

Azaezel:
as discussed in the followup conversation to #500, seems like there's been a bit of command-drift since the writeup we were referencing to attempt consistentcy, so back to treating mac and nix differently on this front, looks like

### DIFF
--- a/Engine/lib/convexDecomp/NvThreadConfig.cpp
+++ b/Engine/lib/convexDecomp/NvThreadConfig.cpp
@@ -133,9 +133,15 @@ void tc_spinloop()
    // Pause would do nothing on the Xbox. Threads are not scheduled.
 #elif defined( _WIN64 )
    YieldProcessor( );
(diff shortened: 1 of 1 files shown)


Azaezel:
Should note there is a concern that ARMs won't properly use the pause fallback, though at present, we're thinkin it's relatively unlikely that it's both using an ARM processor, *and* lacking `_POSIX_PRIORITY_SCHEDULING` so might wanna beat this one up for a few days.

dottools:
For ARM is the instruction is `yield`, however you are suppose to call the OS thread management yield equivalent function instead so that your task will be context swapped out.

The only code that should ever use specialized instructions like this are kernel level, like device drivers and such that have a real reason for using it. Game engines are far too high level to not be using the thread API directly.

Azaezel:
So... consensus?

PR #517, opened 2021-07-31
Update TinyXML to TinyXML2

lukaspj:
This addresses an issue where engine export would fail due to very long doc-strings.
This does not address the fact that doc-strings are sometimes corrupted.

Apart from immediate fixes, TinyXML2 should also be more efficient and focused on game engines it seems.

The focus here has been to do a drop-in replacement, no significant refactoring/fixing has been done in the usage of TinyXML


lukaspj:
I have only had time to test simple xml and engine exports.
Would make sense if someone with insight into those systems could test the other parts this touches, like collada export/import

Azaezel:
seemed to integrate fairly well, though I will note the newer version seems to have converted a significant chunk of what were carriage returned one liners into spaces when it came to variable vectors. so as an example

```

        <EngineFunction
            name="cycleResources"
            returnType="void"
            symbol="fncycleResources"
            isCallback="0"
            isVariadic="0"
            docs="">
            <arguments />
```
            became
            
```
        <EngineFunction name="cycleResources" returnType="void" symbol="fncycleResources" isCallback="0" isVariadic="0" docs="">
            <arguments/>
        </EngineFunction>
        <EngineFunction name="isKoreanBuild" returnType="bool" symbol="fnisKoreanBuild" isCallback="0" isVariadic="0" docs="isKoreanBuild()">
            <arguments/>
```

lukaspj:
Yeah that's a bit of a shame, I can't really find any easy way to add that back, but nothing is impossible

Azaezel:
no worries, just means we'll have to be really sure we're not pulling any explicit readline calls.

Azaezel:
so we're clear: seems to check out so far, just holding off on rolling this one in till closer to when we throw over the major asset fork we're beating up since that's heavily xml based and serves as our primary test case.

PR #518, opened 2021-07-31
Avoid corruption of asset property docstrings

lukaspj:
I don't know exactly why, but the weird `String(String(),String()).c_str()` construct ended up with corrupted docstrings.
So I converted it to plain old C macro concatenation

### DIFF
--- a/Engine/source/T3D/assets/ImageAsset.h
+++ b/Engine/source/T3D/assets/ImageAsset.h
@@ -154,7 +154,7 @@ public:
    static void consoleInit();
 };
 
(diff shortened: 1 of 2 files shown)


lukaspj:
From Discord:
This could alternatively be implemented with this instead: `#define assetText(x,suff) (new std::string(std::string(#x) + std::string(#suff)))->c_str()`

lukaspj:
I pushed the second option. The original was:
`#define assetText(x,suff) #x #suff`

lukaspj:
And we're back to the original solution with "#x #suff"

Azaezel:
addressed in https://github.com/TorqueGameEngines/Torque3D/pull/520. thanks again.

PR #525, opened 2021-08-10
Updates openAL to bypass the cmake failure that the older library causes

Bloodknight:


lukaspj:
Considering that this is included in 4.0:
https://github.com/TorqueGameEngines/Torque3D/blob/development/Engine/lib/openal-soft/ChangeLog#L1 (1.21.1 so even newer there)
And it's covered by 3.11 branch:
https://github.com/Areloch/Torque3D/blob/Release_3_11_Test/Engine/lib/openal-soft/ChangeLog#L1 (1.21.1 so even newer there as well)

I reckon we can close this one?

PR #529, opened 2021-08-12
get the splash screen on linux to stop corrupting the main window

Azaezel:


Azaezel:
original attribution: ragora

PR #537, opened 2021-08-17
* Feature: Implement CI using Github actions.

Ragora:
This merge enables Github itself as the CI for the project.

### DIFF
--- a/.github/workflows/ubuntu-gcc.yml
+++ b/.github/workflows/ubuntu-gcc.yml
@@ -1,0 +1,44 @@
+name: Ubuntu GCC
+
+# Controls when the action will run.
+on:
+    # Triggers the workflow on push or pull request events but only for the develop branch
+    push:
+        branches: [Preview4_0]
+    pull_request:
+        branches: [Preview4_0]
+
+    # Allows you to run this workflow manually from the Actions tab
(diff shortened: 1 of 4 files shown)


.github/workflows/ubuntu-gcc.yml
```diff
@@ -0,0 +1,44 @@
...
+
+        # Steps represent a sequence of tasks that will be executed as part of the job
+        steps:
+            # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+            - uses: actions/checkout@v2
+
+            - uses: lukka/get-cmake@latest
+
+            - name: Install Ubuntu Dependencies
+              run: sudo apt-get install -y libgtk-3-dev libopenal-dev libxxf86vm-dev
```
lukaspj, line 37: Is this enough? Seems like a way shorter list than the one in the Travis file?

Ragora: It seems to build for Ubuntu, but I can add further dependencies if they're needed: https://github.com/Ragora/Torque3D/actions

lukaspj: It might be a difference in build vs run, I think I experienced that when running T3D in Docker

PR #538, opened 2021-08-17
Update main.css

aasim-syed:
i have made the website more aestheic.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -7,7 +7,7 @@ body {
  */
 
 .table.borderless > tbody > tr > td {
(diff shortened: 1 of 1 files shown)


lukaspj:
Afaik that site is not used

Areloch:
Yeah, as lukas noted, the main site now utilizes a dedicated CMS platform as opposed to the original gh-pages there.
That said, even so, I wanted to say thank you @aasim-syed for taking the time to put a little love into it. I very much appreciate your effort! :)

PR #541, opened 2021-08-21
stringtable takes the first case-combo it runs across

Azaezel:


dottools:
Ummm... Why the case changes to lowercase when everything else is Camel case?

Azaezel:
it was actively fouling up material.cs reference upgrades leading to an inability for nix machines to handle them

PR #544, opened 2021-08-22
[PLEASE TEST BEFORE MERGE] Mac font stuff

JeffProgrammer:
So..I rewrote the font renderer for MacOS. The old one was absolutely hideous on my retina MacBook Pro when looking at it. However, while this PR improves it, should it be made better. 

Any Mac users out there who want to test this and see what it looks like on their Macs before we merge into 4.0?


Azaezel:
seems to have worked thusfar, so likely go in tonight or tomorrow unless anyone's noted side effects

PR #557, opened 2021-08-31
sunlight is inherently more bright than other cases.

Azaezel:
adds a multiplier for higher range variance using the same input value

### DIFF
--- a/Engine/source/environment/scatterSky.cpp
+++ b/Engine/source/environment/scatterSky.cpp
@@ -258,7 +258,7 @@ void ScatterSky::_conformLights()
       lightDirection = -mMoonLightDir;
 
    mLight->setDirection( -lightDirection );
-   mLight->setBrightness( brightness * mBrightness );
+   mLight->setBrightness( brightness * mBrightness *2 );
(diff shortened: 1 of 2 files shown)


lukaspj:
I think this is a bit hacky, I understand why but I'd personally prefer if there were a more transparent solution.
Ideally the brightness range should just change based on the type of light source, since it sound like the issue here is that the input slider GUI does not have a big enough range?

Azaezel:
More a case of on the one hand the values at present being largely arbitrary, and on the other hand avoiding in shader per-pixel calcs for a given light type.

dottools:
Could it not be a configurable value from a variable like  `gWorldAmbientLightFactor` instead of using a magic number `2`?
I forget what Torque's global variable style looks like, so my example global variable is prefixed with a 'g'.

Azaezel:
Sure, but you'd still want to apply that once to the value passed to the shader, not *within* it, multiple times. So the general categories of options are:
A) Multiplier or Sum
B) Higher initial value
Could see going either route, honestly.

PR #559, opened 2021-09-02
* Workaround: GCC Release mode Runtime Errors

Ragora:
This merge works around apparent compiler bugs in GCC causing an application crash right at startup.

### DIFF
--- a/Engine/source/core/util/str.cpp
+++ b/Engine/source/core/util/str.cpp
@@ -284,8 +284,8 @@ class String::StringData : protected StringDataImpl
             delete [] mUTF16;
       }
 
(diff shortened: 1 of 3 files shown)


jwakely:
This is a bug in your code, not GCC.

https://gcc.gnu.org/gcc-6/porting_to.html#flifetime-dse

Azaezel:
Wouldn't be the first time. Re-reviewing.

thesamesam:
For completeness: this seems to have been addressed by https://github.com/TorqueGameEngines/Torque3D/issues/810 and https://github.com/TorqueGameEngines/Torque3D/pull/812.

PR #580, opened 2021-09-16
* [Text Pad] BugFix: Correct a local variable reference

Ragora:
This corrects a local variable reference causing the GUI Editor (and other editors using the Text Pad) to crash.


Ragora:
Seemingly was just me being nuts.

PR #585, opened 2021-09-20
Sound asset implements

marauder2k7:


Engine/source/T3D/fx/lightning.h
```diff
@@ -86,6 +92,22 @@ class LightningData : public GameBaseData
...
+
+   SFXProfile* getThunderProfile(U32 id)
+   {
+      if (mThunderSoundAsset[id] != NULL)
+         return mThunderSoundAsset[id]->getSfxProfile();
+      else
+         return NULL;
+   }
+
+   SFXProfile* getSFXProfile() {
```
dottools, line 104: Uh oh. `One true brace` coding style leaked here...

Also it is better coding style for functions that return values to always have a non-conditional return statement at the end of the function, instead of nested under an else, like so:

```cpp
SFXProfile* getSFXProfile()
{
   if (mStrikeSoundAsset.notNull())
      return mStrikeSoundAsset->getSfxProfile();

   return NULL;
}
```

That way the intent is more clear. If mStrikeSoundAsset is not NULL return it. At end of function it just returns NULL as in failure to satisfy request, operation failed, etc..

Engine/source/T3D/fx/splash.h
```diff
@@ -91,8 +92,19 @@ class SplashData : public GameBaseData
...
-   AudioProfile*           soundProfile;
-   S32                     soundProfileId;
+   //AudioProfile*           soundProfile;
+   //S32                     soundProfileId;
+
+   DECLARE_SOUNDASSET(SplashData, Sound);
+   DECLARE_SOUNDASSET_SETGET(SplashData, Sound);
+
+   /// this should probably be added as a function higher up to stop repeats.
+   SFXProfile* getSFXProfile() {
```
dottools, line 102: Eww it happened again...

Engine/source/T3D/projectile.cpp
```diff
@@ -368,9 +367,8 @@ bool ProjectileData::preload(bool server, String &errorStr)
          if (Sim::findObject(decalId, decal) == false)
             Con::errorf(ConsoleLogEntry::General, "ProjectileData::preload: Invalid packet, bad datablockId(decal): %d", decalId);
 
-      String sfxErrorStr;
-      if( !sfxResolve( &sound, sfxErrorStr ) )
-         Con::errorf(ConsoleLogEntry::General, "ProjectileData::preload: Invalid packet: %s", sfxErrorStr.c_str());
+      if( !getProjectileSound() )
```
dottools, line 370: Feel free to delete the erroneous/unnecessary spaces inside parenthesis (right of `(` and left of `)`). Optionally space between keyword and expression (except for operators, such as like `sizeof()`, casts)
example:
```cpp
if( !getProjectileSound() )
```
to
```cpp
if (!getProjectileSound())
```

Engine/source/T3D/projectile.h
```diff
@@ -115,7 +116,14 @@ class ProjectileData : public GameBaseData
    DecalData *decal;                   // (impact) Decal Datablock
    S32 decalId;                        // (impact) Decal ID
 
-   SFXTrack* sound;                    // Projectile Sound
+   DECLARE_SOUNDASSET(ProjectileData, ProjectileSound);
+   DECLARE_SOUNDASSET_SETGET(ProjectileData, ProjectileSound);
+   SFXProfile* getSFXProfile() {
```
dottools, line 122: And again :(

Engine/source/T3D/vehicles/wheeledVehicle.cpp
```diff
@@ -301,7 +301,7 @@ WheeledVehicleData::WheeledVehicleData()
    wheelCount = 0;
    dMemset(&wheel, 0, sizeof(wheel));
    for (S32 i = 0; i < MaxSounds; i++)
-      sound[i] = 0;
+      INIT_SOUNDASSET_ARRAY(WheeledVehicleSounds, i);
```
dottools, line 304: Danger! Do not trust preprocessor to handle code body braces for you in `if ()`, `while ()`, `for ()`, etc.. statements as it is known to be [broken for various compilers and can lead to bugs](https://stackoverflow.com/questions/154136/why-use-apparently-meaningless-do-while-and-if-else-statements-in-macros). Always explicitly put macros inside of code body braces:
```cpp
for (S32 i = 0; i < MaxSounds; i++)
{
   INIT_SOUNDASSET_ARRAY(WheeledVehicleSounds, i);
}
```

Even though `INIT_SOUNDASSET_ARRAY` macro definition is code body braced doesn't mean all preprocessors will handle it well. And for others the classic annoying unnecessary trailing ; warning message.

Engine/source/T3D/fx/lightning.h
marauder2k7: copied from the way it was done before

Engine/source/T3D/fx/splash.h
marauder2k7: copied from the way it was done before

Engine/source/T3D/projectile.cpp
marauder2k7: copied from the way it was done before

Engine/source/T3D/projectile.h
marauder2k7: copied from the way it was done before

Engine/source/T3D/vehicles/wheeledVehicle.cpp
marauder2k7: copied from the way it was done before

Engine/source/T3D/fx/lightning.h
marauder2k7: https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Engine/source/T3D/fx/precipitation.h#L76

Engine/source/T3D/fx/splash.h
marauder2k7: https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Engine/source/T3D/fx/precipitation.h#L76

Engine/source/T3D/projectile.h
marauder2k7: https://github.com/TorqueGameEngines/Torque3D/blob/Preview4_0/Engine/source/T3D/fx/precipitation.h#L76

Azaezel:
git mangled this PR, so we followed up via #616

PR #608, opened 2021-10-08
BugFix: Correct MSVC Compiler Warnings

Ragora:
This merge attempts to address compiler warnings emitted by MSVC.

### DIFF
--- a/Engine/source/platform/platformNet.cpp
+++ b/Engine/source/platform/platformNet.cpp
@@ -95,36 +95,6 @@ typedef int SOCKET;
 
 #endif
 
-#if defined(TORQUE_USE_WINSOCK)
-static const char* strerror_wsa( S32 code )
-{
-   switch( code )
-   {
-#define E( name ) case name: return #name;
-      E( WSANOTINITIALISED );
-      E( WSAENETDOWN );
-      E( WSAEADDRINUSE );
-      E( WSAEINPROGRESS );
-      E( WSAEALREADY );
-      E( WSAEADDRNOTAVAIL );
-      E( WSAEAFNOSUPPORT );
-      E( WSAEFAULT );
-      E( WSAEINVAL );
-      E( WSAEISCONN );
-      E( WSAENETUNREACH );
-      E( WSAEHOSTUNREACH );
-      E( WSAENOBUFS );
-      E( WSAENOTSOCK );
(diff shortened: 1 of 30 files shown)


Engine/source/platform/platformNet.cpp
```diff
@@ -1262,7 +1233,7 @@ void Net::process()
          {
             // try to connect
             out_h_addr.port = currentSock->remotePort;
-            const sockaddr *ai_addr = NULL;
+            const sockaddr* ai_addr = NULL;
```
dottools, line 1236: Oh no. This is improper as it is the variable that is a pointer, not a pointer type. Very bad habit when declaring variables.
The original code was proper.

Engine/source/platform/platformNet.cpp
```diff
@@ -1304,7 +1275,7 @@ void Net::process()
             else
             {
                Con::errorf("Error connecting to %s: Invalid Protocol",
-               currentSock->remoteAddr);
+                  currentSock->remoteAddr);
```
dottools, line 1278: ~~What's going on with these random indents? Same on line 1293.~~
Nevermind on these two changes, this is actually correct. This is what happens when I look away from the screen for a moment and then return my eyes to wrong line. Sorry.

Engine/source/platform/platformNet.cpp
Ragora: The same argument can be made for it being a pointer type which is exactly the case at my previous job. There needs to be a programming style guide to agree upon for cases like this.

dottools: It has to do with the language. This applies for both C and C++:
```cpp
const sockaddr* a1, b2, c3;
```

`a1` is a pointer variable to `const sockaddr` memory, while `b2` and `c3` are `const sockaddr` variables.
Which is why it is improper to have the asterisk next to the data type because it is not apart of the data type declaration.

Ragora: That honestly on its own makes no sense, they *all* should be a sockaddr pointer I'd say, so I'd argue you shouldn't even be declaring them like that.

JeffProgrammer: the dumbness of C and C++. :x

PR #613, opened 2021-10-09
followups to #582

Azaezel:


Azaezel:
There *will* be more work done on a cleanup front, but this stops most of the errorspam and a few crashes, so using the savepoint, as it were.

PR #627, opened 2021-10-16
sound asset followups

Azaezel:


Azaezel:
TODO list 16OCT2021:
afxAudioBank-
this serves as effectively a tool for dropping a directory in whole cloth. will need to determine how we wish to apply assetification principles here

guiTextEditCtrl and guiTypes fall into a circular dependency we'll need to resolve

PR #637, opened 2021-10-26
blatantly ganked from T2D; adds rotation as an option for drawbitmap

Azaezel:
demonstrates a bit of usage in guibitmapctrl. (do note, gui elements may need followup via altering thier clip rect depending on how folks want to end up using this)


JeffProgrammer (approved):

dottools:
I know this was copied from T2D, but we don't normally use type prefixes in argument names. `fSpin`

Even calling it `spin` is a misnomer as it isn't spinning (which implies animation) when it is just a rotation angle.
So the argument could be named `rotation` or `angle` instead.

PR #652, opened 2021-11-02
BugFix: Correct a crash in the variable inspector

Ragora:
This PR addresses a potential crash in the variable inspector caused by a user-after-free memory error. The problem arises from the variable inspector holding pointers into tVector managed memory which may be freed at any point in time later in execution (Eg. due to a tVector resize), causing a dangling pointer into this memory space.

So we allocate these entries individually on the heap and hold pointers in the tVector to them.


dottools:
Honestly I don't see any fixes here. All you've done is changed `mFields` to manual resource memory management instead of letting the `Vector` STL container do it for us.

Ragora:
The fact that TVector was managing it for us is the problem. We were maintaining pointers into a buffer that can be reallocated at any point in time due to a push_back call and invalidate all pointers into said buffer which is what led to the crash. If this is not an acceptable fix then the code simply needs written.

I guess I should have included a reproduction to show it does do something:

Screen Shot 2021-11-02 at 10 08 48 PM

Repeatedly clicking "DefaultImportConfig" prior to the patch would cause the engine to crash.

dottools:
Hmmmm. I suppose that the internal `Vector` memory manager could end up calling `realloc()` when its existing memory allocation becomes fully occupied which would invalidate the existing pointers to the container's members.

PR #674, opened 2021-11-16
Adjustment: Generalization of platformX86UNIX to platformPOSIX

Ragora:
This PR generalizes most of platformX86UNIX to platformPOSIX. It also enables the enableWinConsole(1); command to work on MacOS.


dottools:
This looks like it was only tested on Mac. Platform X86UNIX was also meant for Linux.
I also noticed that you deleted checking for dedicated server builds of which will no longer work as intended.

Ragora:
I retested to verify my sanity on Linux and it runs. I then double checked if anything used the class (by removing the class fully and recompiling) with no problems. I then triple checked it with a grep that turned up nothing. So unless we want to keep the class for some reason I should have removed it entirely.

PR #676, opened 2021-11-17
Optionally allow to treat script assert as warning

JeffProgrammer:
This commit allows us to treat variable use before assign errors and local variables inside of the global scope as warnings instead of asserts. This will allow for easier porting of legacy scripts. It is highly recommended use this as an aid to port scripts, but can be used in production if needbe.


JeffProgrammer:
Before merge, need to test eval.

PR #682, opened 2021-11-23
rip out sideprojection

Azaezel:
the shader feature, not the behaviour that is now automatic


lukaspj:
Eh, isn't this removing both features and behavior?

Azaezel:
from the results, seems not, though do please throw some extra samples at it

PR #696, opened 2021-12-04
BugFix: Correct compilation for MacOS

Ragora:
This PR addresses compilation errors on MacOS. It uses C++17 on the primary engine executable and C++11 on all libraries we compile ourselves - seemingly not specifying the C++ standard explicitly for libraries caused strange compilation errors in Assimp, for example.


rextimmy:
Sorry this is my bad, I did test it with gcc but not clang. I wonder if it would be better to let each library in the Tools/CMake/libraries directory set what cxx version it requires?

PR #700, opened 2021-12-08
clean up normalmap results for forward lit mats

Azaezel:


Azaezel:
killing this as it was just a remindernote/temp workaround.

PR #704, opened 2021-12-13
A clean implementation of Lukas' Fix side projection #684 PR with Az's addendum fix rolled in

Areloch:


lukaspj:
Note that iirc this still has some issues with extreme cases that I haven't worked out.

The normal is computed.. not wrong but in a way that's sensitive to edge cases.

Say you have 3 vertices next to each other, ignoring the 3rd dimension here: (-1; 0), (0; 0) and (1; 100).

Then we want to compute the normal for the middle one. Intuitively, this normal should point at a 45 degree angle right?
Actually we would probably prefer having two normals, one for each "side" of the vertex, but alas we have one.

What actually happens is that the magnitude of vector between mid point and right point is so big, that it cancels out the vector between left and mid, making the normal point straight left (1; 0).

Imo the magnitude of the vector shouldn't matter, only the direction, so I think the connotation of the normals for the terrain should be changed

PR #706, opened 2021-12-14
adress mounted turret desync

Azaezel:
directly fix a reported desync for turrents mounted to other objects when they ghost in and out due to being beyond visibledistance from a given client using the reported workaround from Atomic Walrus
also goes ahead and retriggers setMaskBits(MountedMask) for SceneObject::setTransform as a more generic approach

### DIFF
--- a/Engine/source/T3D/turret/turretShape.cpp
+++ b/Engine/source/T3D/turret/turretShape.cpp
@@ -629,6 +629,9 @@ void TurretShape::processTick(const Move* move)
    if (!isGhost())
       updateAnimation(TickSec);
 
+   if (isMounted()) { // this sets PositionMask at the Item level, forcing a network update and fixing scoping/de-ghosting
+      setVelocity(mMount.object->getVelocity());
+   }
    updateMove(move);
 }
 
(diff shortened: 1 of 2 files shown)


Engine/source/T3D/turret/turretShape.cpp
```diff
@@ -629,6 +629,9 @@ void TurretShape::processTick(const Move* move)
    if (!isGhost())
       updateAnimation(TickSec);
 
+   if (isMounted()) { // this sets PositionMask at the Item level, forcing a network update and fixing scoping/de-ghosting
```
dottools, line 632: Yuck formatting....

```cpp
if (isMounted())
{
   // set PositionMask at the Item level to force a network update in order to fix scoping/de-ghosting
```

Much better and to the point.

Engine/source/scene/sceneObject.cpp
```diff
@@ -433,6 +433,9 @@ void SceneObject::setTransform( const MatrixF& mat )
    if( mSceneManager != NULL )
       mSceneManager->notifyObjectDirty( this );
 
+   if (isMounted()) //try to ensure mounts stay synced
+      setMaskBits(MountedMask);
```
dottools, line 437:
```cpp
if (isMounted())
{
   // network sync mounts
   setMaskBits(MountedMask);
}
```

PR #724, opened 2022-02-01
WIP: Shape constructor asset pack

Areloch:
Packs TSShapeConstructors into asset definition taml file 
Also updates importers to facilitate this behavior

Requires #723


Azaezel:
to be revisited post 4.0 as this would purely be an editor-result file that goes unshipped with a finished product

PR #726, opened 2022-02-14
Improves logical checks for the default value so it's more sane and stable

Areloch:
Improves logical checks for the default value so it's more sane and stable

Also Allows creation of polyhedrons via constructor with a origin and vector format, allowing to have default polyhedron values on triggers and physical zones

### DIFF
--- a/Engine/source/T3D/physicalZone.cpp
+++ b/Engine/source/T3D/physicalZone.cpp
@@ -114,6 +114,13 @@ PhysicalZone::PhysicalZone()
    force_mag = 0.0f;
    orient_force = false;
    fade_amt = 1.0f;
+
+   //Default up a basic square
+   Point3F vecs[3] = { Point3F(1.0, 0.0, 0.0),
+      Point3F(0.0, -1.0, 0.0),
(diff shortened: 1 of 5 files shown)


Engine/source/T3D/physicalZone.cpp
```diff
@@ -114,6 +114,13 @@ PhysicalZone::PhysicalZone()
    force_mag = 0.0f;
    orient_force = false;
    fade_amt = 1.0f;
+
+   //Default up a basic square
+   Point3F vecs[3] = { Point3F(1.0, 0.0, 0.0),
```
dottools, line 119: I know this is nitpicky, but after the past year at work being a code janitor cleaning up everybody else's mess I've developed hatred towards ugly code structure and random alphabet soup variable naming conventions. Can we please make things like this easier on reviewers' eyes like so?:
```cpp
Point3F vecs[3] =
{
   Point3F(1.0,  0.0, 0.0),
   Point3F(0.0, -1.0, 0.0),
   Point3F(0.0,  0.0, 1.0)
};

```

PR #729, opened 2022-02-15
template mixins need this-> specified

Azaezel:


lukaspj:
Huh, why is this necessary?

Azaezel:
Unlike more traditional inheritance, using the template route gets it confused when it comes to member variables for the resulting class(es) if you aren't explicit. Brought to light on the nix side.

dottools:
Doesn't help when C++ committee keeps changing template behavior on every other new standard. They've become UX designers, can't stop changing shit for the sake of change.

PR #749, opened 2022-03-20
Compute Shaders

marauder2k7:
-Adds an interface for compute shaders.

Still very early bones, need a way to set the number of threads required for glDispatchCompute


marauder2k7:
this is now outdated, newer branch with all expanded shaders

PR #751, opened 2022-03-23
update sdl to https://github.com/libsdl-org/SDL 22March 2022

Azaezel:
precise sync point: https://github.com/libsdl-org/SDL/commit/006aafabcf76545e3712529b28c8bcec862c4cc5


Azaezel:
just to keep folks in the loop on this one:
holding off on an updated follow-up until https://github.com/libsdl-org/SDL/milestones is locked in

Azaezel:
so given prior to this certain unix varieties were unable to compile, and we do want to have a revised version of the project manager and alpha builds up prior to the newly revised sdl timeline of end of april, going ahead and rolling this in after a few folks 'nix tested

PR #779, opened 2022-04-28
Make the Console Sane Again

JeffProgrammer:


dottools:
That was some ulgy hacks.

Issue #782, opened 2022-05-16
Copy/Paste on macOS


CouleeApps:
Copy/paste currently does not work in gui text boxes on macOS. Not sure if this is a bug in the mac input handling or the gui boxes, but (i assume) if it works on windows it needs to also work on mac.

PR #792, opened 2022-05-26
BugFix: Correct MacOS not responding to various hotkeys

Ragora:
This PR addresses https://github.com/TorqueGameEngines/Torque3D/issues/782 with MacOS not responding to Command+C, Command+A, Command+V and Command+X in text input controls such as the console.


Ragora:
Note upon further inspection it looks like a more intelligent approach is needed if we don't want to just treat command like ctrl on other platforms.

PR #813, opened 2022-06-12
Fix Misc ConvexShape Tooling Issues

Areloch:
- Fixes bounds scaling issue making the object box in the editor be double the convex's size
- Fixes loading of modified surfaceTextures by properly inserting into stringtable
- Fixes display of the active and default material previews in the ConvexShape editor
- Adds buttons to hollow and recenter selected convex to tool window

### DIFF
--- a/Engine/source/gui/worldEditor/guiConvexShapeEditorCtrl.cpp
+++ b/Engine/source/gui/worldEditor/guiConvexShapeEditorCtrl.cpp
@@ -1075,7 +1075,7 @@ void GuiConvexEditorCtrl::renderScene(const RectI & updateRect)
          Point3F boxPos = objBox.getCenter();
          objMat.mulP( boxPos );
          
-         drawer->drawObjectBox( desc, objBox.getExtents(), boxPos, objMat, ColorI::WHITE );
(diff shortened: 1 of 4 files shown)


Engine/source/gui/worldEditor/guiConvexShapeEditorCtrl.cpp
```diff
@@ -1075,7 +1075,7 @@ void GuiConvexEditorCtrl::renderScene(const RectI & updateRect)
          Point3F boxPos = objBox.getCenter();
          objMat.mulP( boxPos );
          
-         drawer->drawObjectBox( desc, objBox.getExtents(), boxPos, objMat, ColorI::WHITE );
+         drawer->drawObjectBox( desc, objBox.getExtents() / 2, boxPos, objMat, ColorI::WHITE );
```
dottools, line 1078: This doesn't seem right. Need to figure out why the dimensions are twice the size than expected.
I suspect that this couple lines above has something to do with it:
```cpp
objBox.scale( mConvexSEL->getScale() );
```

Azaezel: we'll revisit the /2 when more time allows

PR #819, opened 2022-06-15
you want the higher number, not the lower

Azaezel:


dottools:
Doh! My fault. Every time I see `min` in `std::min()` I think of `at minimum value` instead of `lowest of the two`.
You know what, `min()`/`max()` macros, later become functions, were named wrong the whole time!

PR #852, opened 2022-08-16
WinKey+Shift+L/R Arrow handler.

OTHGMars:
Attempts to constrain canvas when switching between monitors via keyboard shortcut.


Jozor:
Can confirm this works on my multi-monitor 1080p/1440p/1080p setup.

PR #856, opened 2022-08-17
specular code review.

Azaezel:
the CAPTURING flag was being improperly checked for existence instead of state, leading to overbrightness in the precence of metalness. removed hacky workaround, and applied the proper fix.
also foreshortened the roughness range to avoid values outside of f0 going asymptotic.
alternative suggestion for if that portion plays up:
   surface.roughness = pbrProperties.b*0.92f+0.04f;
to
  clamp(gbuffer2.b, 0.04f, 0.96f);
(basically a case of do we support more fine-grained differences or just chuck anything too extreem)


Azaezel:
hat tip @GoldenThumbs for helpin to eyeball the trouble spots

PR #878, opened 2022-09-03
WIP: libcurl integration with HTTPObject

CouleeApps:
Here's a version of libcurl for Torque3D, mostly copied from MBExtender (MIT). Marking this as WIP since I haven't really tested it nor have I investigated the security implications. It worked in a trivial test case though.


Azaezel:
This has been moved up in priority as the utility has been leveraged for the new project manager. Thus far there seems to have been no issues with utilization, but if folks want to give this a second look by end of week, it would be greatly appreciated.

PR #894, opened 2022-09-16
Implements SMAA and a few unaffiliated fixes

Areloch:
- Adjusts handling for httpObject class so that if curl flag is off, files are not included and references are disabled to avoid compilation issues
- Adds SMAA anti aliasing
- Adjusts AA option setting to properly work and facilitate SMAA Added defaults for AA to facilitate SMAA
- Updated brightness and HDR values on Example Level to comply more closely to the Editor default to give better default results - Fixed issue where ShapeEditor wasn't initialized properly when we'd open it from the Asset Browser


JeffProgrammer:
mac seems OK from what I can tell. no shader errors at least.

PR #913, opened 2022-10-19
add backbufferSRGB option (defaults to true) for those postfx that ex…

Azaezel:
…pect to have everything in linear space to begin with

updated smaa to use it


Azaezel:
gonna put a pin in this line of experimentation till we circle back to focus on the next gfx rev

PR #924, opened 2022-11-28
Compilation fixes for C++20

JeffProgrammer:


dottools:
What's with all the usages of `NULL` instead of `nullptr`?

JeffProgrammer:
the files tended to use NULL, so I kept it the same. I guess I could change it.

Azaezel:
> the files tended to use NULL, so I kept it the same. I guess I could change it.

prettymuch just leftovers from the prior U32 wrapping route most folks took. While folks should be aware that the DB memblock does still use that for offset calcs, the lions share of utilities shouldn't really care one way or the other.

PR #932, opened 2022-12-08
adds a mechanism to inject additional steps into mission loading

Azaezel:
leverages the EventManager and ScriptMsgListener() classes to set up a third mission load stage triggered by the following flow: 
function <module>::onLoadMap(%this) starts an execution chain that leads to <module>::finishMapLoad().
each  <module>::finishMapLoad() MUST contain the line
     Core_ClientServer.GetEventManager().postEvent( "mapLoadComplete" );
once all have called back that they have finished their tasks, players finish loading into a hosted mission

### DIFF
--- a/Templates/BaseGame/game/core/clientServer/Core_ClientServer.tscript
+++ b/Templates/BaseGame/game/core/clientServer/Core_ClientServer.tscript
@@ -39,6 +39,32 @@ function Core_ClientServerListener::onMapLoadComplete(%this)
     }
 }
 
+function Core_ClientServerListener::onmapLoadFail(%this, %isFine)
+{   
+    if (%isFine) 
+    {
+        %this.onMapLoadComplete();
+        return;
+    }
+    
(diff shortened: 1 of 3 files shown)


Azaezel:
Will note: Never a real fan of musts, so by all means, open to alternative suggestions

dottools:
You know what this means. Soon we'll need ordered dependency mission loading callbacks. Followed by optional retries on failure. Eventually we'll get to systemd level craziness and we'll just have to curl up and cry ourselves to sleep each night.... ;)

Azaezel:
weeeell... as far as that goes, module order dependency's already a thing. and the retry on a fail bit is one of the reasons for cooking this up to begin with. was working on a procedural generation module with a potential for a failed outcome where it retries. which could take a bit. hence the 'and get back to me when you're done' injection via listener. so already a good 50-75% there on those fronts. Though higher order interleaving is likely beyond scope till 4.1 proper.

PR #943, opened 2022-12-29
emissive  to recivesShadows

Azaezel:
now we've got a glow mask and multiplier, ditch the emissive flag in favor of a proper recivesShadows


AtomicWalrus (commented):
FYI emissive has uses that aren't covered by disabling shadows or glow.

For example, anytime you want to render a display (TV, monitor) you'd want it to display the image without any lighting influence. Especially important if you're rendering GUI's to in-world screens.

You can get a similar effect with "glow," but it will blur the content which isn't always desired.

Not a big deal, I just reverted this in my own code, but I wanted to mention it because I think the logic in this change was that emissive was only used for special effects like water and similar.

PR #958, opened 2023-02-13
Team support

Azaezel:
leverages a U8 for up to 256 teams assigned as masking to shapebase derivatives
also applies teamids and IFF to physical zones, prox mines, and turrets. for the latter, teamID 0 is considered friendly so ai turrets don't go after destructible background objects.


dottools:
Seems like a lot of network overhead if every ShapeBase object has a teamID assigned to it.

Did you consider using SimSets and onClient callbacks for team associated objects like how Tribes did it?

dooderino:
> Seems like a lot of network overhead if every ShapeBase object has a teamID assigned to it.
> 
> Did you consider using SimSets and onClient callbacks for team associated objects like how Tribes did it?

Is there an example on GitHub or elsewhere of how Tribes did it?

Azaezel:
> Seems like a lot of network overhead if every ShapeBase object has a teamID assigned to it.
> 
> Did you consider using SimSets and onClient callbacks for team associated objects like how Tribes did it?

Yes, and in practice, it turned into more of a performance hit than the 8 bits spent on tracking once ai, physics, et al were taken into account. Though I could see reducing that aspect even further if we were willing to reduce the amount of teams for king of the hill. Only reason we send the full U8 is to preserve that cap.

Azaezel:
so as a followup to the raised concerns, if we do end up incorporating this, we're likely to want to incorporate a check at the `ExtendedInfo` filter level to pack the state in at a discrete once and done unless notified otherwise rather than a constant send.

PR #968, opened 2023-02-21
Enables terrain macro maps, adds height blend "hardness" setting

AtomicWalrus:
-Enables macro maps. These are optional, and layers with and without macros can be used together. These are diffuse-only tiled textures that are used similarly to detail maps, but at a longer range and generally higher scale to add mid-range detail. They do not currently use normal or ORM maps, or respond to parallax (same behavior as 3.x)
-Updates macro code to follow the design of detail map code (arrays for layer info, texture resource counter incrementation adjusted for deferred).
-Adds a "blend hardness" variable to terrain materials, and adds controls for this variable to the terrain mat editor. Controls how quickly the detail and macro textures fade away when blending with an adjacent layer. A value of 0 produces the previous stock behavior (soft blend biased by heightmap). Higher values create faster transitions, still biased by heightmap. For extremely sharp transitions, both adjacent layers should be set to high hardness. See this example video: https://www.youtube.com/watch?v=li_56CzSSBQ

### DIFF
--- a/Engine/source/terrain/hlsl/terrFeatureHLSL.cpp
+++ b/Engine/source/terrain/hlsl/terrFeatureHLSL.cpp
@@ -46,7 +46,7 @@ namespace
       FEATUREMGR->registerFeature( MFT_TerrainParallaxMap, new NamedFeatureHLSL( "Terrain Parallax Texture" ) );   
       FEATUREMGR->registerFeature( MFT_TerrainDetailMap, new TerrainDetailMapFeatHLSL );
       FEATUREMGR->registerFeature( MFT_TerrainNormalMap, new TerrainNormalMapFeatHLSL );
-      FEATUREMGR->registerFeature( MFT_TerrainMacroMap, new NamedFeatureHLSL("TerrainMacroMap Deprecated")); // new TerrainMacroMapFeatHLSL);
+      FEATUREMGR->registerFeature( MFT_TerrainMacroMap, new TerrainMacroMapFeatHLSL);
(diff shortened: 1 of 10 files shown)


AtomicWalrus:
If anyone wants these two things separated (macro feature vs. blend hardness) to be considered individually let me know and I'll split it into two PRs.

Engine/source/terrain/hlsl/terrFeatureHLSL.cpp
```diff
@@ -918,13 +953,12 @@ ShaderFeature::Resources TerrainMacroMapFeatHLSL::getResources( const MaterialFe
       // If this is the first detail pass then we 
       // samples from the layer tex.
       res.numTex += 1;
-   }
+      res.numTexReg += 1;
```
lukaspj, line 956: This could use a comment in what we are accounting for

AtomicWalrus: This is basically just me duplicating what the detail feature currently does -- I'm not 100% sure why it was written this way (+1-ing those vars twice), but my understanding is that with deferred rendering we no longer need to count every instance of a detail or macro texture against the registers limit. For detail it just adds 2 to the limit if the terrain has any detail layers at all, so I did the same thing for macro.

If you count every detail or macro map against the reg list, you can only have like 4 of either in a cell before the shader builder will start stripping layers out. It has a limit of 10 for texreg

lukaspj: We switched to a single pass approach so we aren’t limited by texture count, since we pack it into an array.

but you need to count that we add the macromap array, so I’m guessing this is too count that

PR #979, opened 2023-03-04
Small fixes

j-mattsson:
Some small fixes.

### DIFF
--- a/Engine/source/gfx/D3D11/gfxD3D11Device.cpp
+++ b/Engine/source/gfx/D3D11/gfxD3D11Device.cpp
@@ -280,7 +280,7 @@ void GFXD3D11Device::enumerateAdapters(Vector<GFXAdapter*> &adapterList)
       SAFE_DELETE_ARRAY(str);
 
       dStrncpy(toAdd->mName, Description.c_str(), GFXAdapter::MaxAdapterNameLen);
-      dStrncat(toAdd->mName, " (D3D11)", 8);
+      dStrncat(toAdd->mName, " (D3D11)", sizeof(toAdd->mName) - strlen(toAdd->mName) - 1);
 
       IDXGIOutput* pOutput = NULL; 
       HRESULT hr;
--- a/Engine/source/gfx/D3D11/gfxD3D11Device.cpp
+++ b/Engine/source/gfx/D3D11/gfxD3D11Device.cpp
@@ -280,7 +280,7 @@ void GFXD3D11Device::enumerateAdapters(Vector<GFXAdapter*> &adapterList)
       SAFE_DELETE_ARRAY(str);
 
       dStrncpy(toAdd->mName, Description.c_str(), GFXAdapter::MaxAdapterNameLen);
-      dStrncat(toAdd->mName, " (D3D11)", GFXAdapter::MaxAdapterNameLen);
+      dStrncat(toAdd->mName, " (D3D11)", 8);
 
       IDXGIOutput* pOutput = NULL; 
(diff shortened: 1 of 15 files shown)


Azaezel:
looks good on a read through. will beat this up over the weekend and see about rolling it in Monday or Wednesday

Engine/source/gfx/D3D11/gfxD3D11Device.cpp
```diff
@@ -280,7 +280,7 @@ void GFXD3D11Device::enumerateAdapters(Vector<GFXAdapter*> &adapterList)
       SAFE_DELETE_ARRAY(str);
 
       dStrncpy(toAdd->mName, Description.c_str(), GFXAdapter::MaxAdapterNameLen);
-      dStrncat(toAdd->mName, " (D3D11)", GFXAdapter::MaxAdapterNameLen);
+      dStrncat(toAdd->mName, " (D3D11)", 8);
```
dottools: This is incorrect. `dStrncat` is a wrapper for `strncat` where third argument is `num` for [Maximum number of characters to be appended](https://cplusplus.com/reference/cstring/strncat/).

Engine/source/gfx/bitmap/ddsFile.cpp
```diff
@@ -178,7 +178,7 @@ U32 DDSFile::getSizeInBytes( GFXFormat format, U32 height, U32 width, U32 mipLev
       "DDSFile::getSizeInBytes - Must be a Block Compression format!" );
 
    // From the directX docs:
-   // max(1, width � 4) x max(1, height � 4) x 8(DXT1) or 16(DXT2-5)
+   // max(1, width ÷ 4) x max(1, height ÷ 4) x 8(DXT1) or 16(DXT2-5)
```
dottools: Your code editor isn't handling unicode properly and has mangled lines 181 and 126.

Engine/source/gfx/D3D11/gfxD3D11Device.cpp
Azaezel: errr `" (D3D11)"` *is* 8 chars?

Areloch: It's a correct value, I think it's mostly a question of from a code standards standpoint, is the hard number value better than the util variable that holds the same?

dottools: All newer functions that have `n` inserted in the middle of the typical string functions, such as `strcat` -> `strncat` in this case, num argument mean what is the max number of characters that destination character buffer may hold.

In this case you're not telling strncat (*dStrncat wrapper for*) the length of `" (D3D11)"`, you're telling it the capacity of `toAdd->mName`.

I believe you guys are confusing these functions with how `memcpy` works.

Linux manpage explains it much better than the awful choice of words from C standard: https://linux.die.net/man/3/strncat

dottools: Actually I re-read the implementation and come to realize what it is actually doing. Gah I really need to stop trying to help when I'm mentally tired.

This article explains it better than I can after I figured out how it should be used: https://codeql.github.com/codeql-query-help/cpp/cpp-unsafe-strncat/

Anyway, the obvious original intention was to tell the string concatenate the character capacity of `toAdd->mName` to avoid a buffer overflow. Of which doing `dStrncat(toAdd->mName, " (D3D11)", 8);` absolutely does not prevent buffer overflow.

PR #988, opened 2023-03-11
fix getDirectoryList

Azaezel:
used dStrcat (note: not dStrncat, so there the buffer destination size *is* the third var)


dottools:
`sizeof` is your friend, cause magic numbers are bad:
```cpp
dStrcat(fullpath, "/\0", sizeof(fullpath));
```

PR #1004, opened 2023-04-03
Console Refactor

lukaspj:
This PR doesn't really accomplish anything in itself.
The aim is that it should be the first step on the road to cleaning up the console by carving out some of its responsibilities.

Right now the Console (more or less) serves as:
Central logging hub
The in-game Console
A centralised scripting hub with an abstract representation of variables, functions and classes
A TorqueScript runtime
Utility dumpster for various things, like script file path resolution

I will try to clean this up in multiple steps, this first one is about migrating everything explicitly related to the scripting engine out into its own folder/namespace.

The goal of the clean up is to make it easier to develop and maintain the console and its components such as TorqueScript.


dottools:
This is going to be interesting.

I was thinking when I saw this:
```cpp
void GameObjectAsset::initializeAsset()
{
...
   if (Con::isScriptFile(mScriptPath))
      TorqueScript::getRuntime()->executeFile(mScriptPath, false, false);
...
```

Probably want an equivalent to support multiple unique scripting language runtimes:
```cpp
   // lookup script runtime associated with script file type
   Con::ScriptRunTime &scriptRuntime = Con::getScriptRuntime(mScriptPath);

   // determine if valid script runtime was found
   if (Con::isRuntime(scriptRuntime))
   {
      // execute script file with looked up script runtime
      scriptRuntime.executeFile(mScriptPath, false, false);
   }
```

Although this idea would imply multiple scripting engines could run simultaneously. ¯\\\_(ツ)\_/¯

lukaspj:
> This is going to be interesting.
> 
> I was thinking when I saw this:
> 
> ```c++
> void GameObjectAsset::initializeAsset()
> {
> ...
>    if (Con::isScriptFile(mScriptPath))
>       TorqueScript::getRuntime()->executeFile(mScriptPath, false, false);
> ...
> ```
> 
> Probably want an equivalent to support multiple unique scripting language runtimes:
> 
> ```c++
>    // lookup script runtime associated with script file type
>    Con::ScriptRunTime &scriptRuntime = Con::getScriptRuntime(mScriptPath);
> 
>    // determine if valid script runtime was found
>    if (Con::isRuntime(scriptRuntime))
>    {
>       // execute script file with looked up script runtime
>       scriptRuntime.executeFile(mScriptPath, false, false);
>    }
> ```
> 
> Although this idea would imply multiple scripting engines could run simultaneously. ¯\_(ツ)_/¯

I reverted this to old `Con::executeFile` interface to make the interface changes smaller and re-use the abstractions we already have.
Although the goal is absolutely to have multiple scripting engines running simultaneously. It would be nice if someone re-did TorqueScript or compiled it to e.g. WebAssembly down the road, and it will be nice being able to gradually adopt another scripting engine or e.g. use another one in release builds.

PR #1005, opened 2023-04-06
Fix Unit Testing Suite

lukaspj:
All tests pass on my machine, which, if nothing else, is better than them not compiling at all.

This also adds a _very_ rudimentary setup to allow TorqueScript tests to be reported in the GTest suite.

Hooking up TorqueScript to the suite should be expanded on more in a different PR, this just paves the path for further developments in that space.


lukaspj:
The test report publishing workflow requires elevated permissions that won't be possible until the PR is merged unfortunately. So I don't think we can have insight into the results of that work in this PR.

Edit: This is the action: https://github.com/mikepenz/action-junit-report 
Examples and usage is available on their page.

PR #1011, opened 2023-04-15
Reverse depth & 32F buffer format

AtomicWalrus:
-Adds reversed depth projection model, significantly increasing depth buffer effective resolution. Same model used in modern Unity & Unreal Engines.
-Adds 32F depth 8U stencil format GFXFormatD32FS8X24 (following DX naming conventions). Note this is a 64-bit format, and likely not suitable for mobile platforms. Revert to GFXFormatD24S8 in renderManager.tscript for mobile & "ancient" platforms. 
-Disables OpenGL-specific projection matrix code, as OGL now uses the same depth value range (1-0) as D3D.
-Corrects alignment of texture format details array.


JeffProgrammer:
Mac does not support glClipControl. this will break mac support. glClipControl is from OpenGL 4.5

AtomicWalrus:
Added a conditional workaround for MacOS.

Unfortunately it's only partially equivalent, so while it's functional (near/far planes are correct) it won't produce the same improvement to depth performance. Looks a bit better than stock, but not by much.

PR #1014, opened 2023-04-27
add safeties for enum math across define boundaries

Azaezel:


Azaezel:
for clarity, this is to head off https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2139r1.html#3.1

PR #1015, opened 2023-04-27
fix warn reports for buffer over-runs

Azaezel:
also misc uninitialized vars

### DIFF
--- a/Engine/source/console/console.cpp
+++ b/Engine/source/console/console.cpp
@@ -1549,22 +1549,23 @@ ConsoleValue evaluatef(const char* string, ...)
 ConsoleValue _internalExecute(S32 argc, ConsoleValue argv[])
 {
    StringTableEntry funcName = StringTable->insert(argv[0].getString());
-
-   const char** argv_str = static_cast<const char**>(malloc((argc - 1) * sizeof(char *)));
-   for (int i = 0; i < argc - 1; i++)
+   if (argc > 1)
    {
-      argv_str[i] = argv[i + 1].getString();
+      const char** argv_str = static_cast<const char**>(malloc((argc - 1) * sizeof(char*)));
+      for (int i = 0; i < argc - 1; i++)
+      {
+         argv_str[i] = argv[i + 1].getString();
+      }
+      bool result;
+      const char* methodRes = CInterface::CallFunction(NULL, funcName, argv_str, argc - 1, &result);
+      free(argv_str);
+      if (result)
+      {
+         ConsoleValue ret;
+         ret.setString(methodRes);
+         return std::move(ret);
+      }
    }
-   bool result;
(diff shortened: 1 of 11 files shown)


Engine/source/console/consoleObject.h
```diff
@@ -863,7 +863,7 @@ class ConsoleObject : public EngineObject
 public:
 
    /// Get the classname from a class tag.
-   static const char* lookupClassName(const U32 in_classTag);
+   static const char* lookupClassName(const U32 in_classTag) {};
```
dottools: How is this even compiling, no return value for non-void return type is not allowed in C++.

Engine/source/console/console.cpp
```diff
@@ -1636,22 +1637,24 @@ static ConsoleValue _internalExecute(SimObject *object, S32 argc, ConsoleValue a
...
-
-   const char** argv_str = static_cast<const char**>(malloc((argc - 2) * sizeof(char *)));
-   for (int i = 0; i < argc - 2; i++)
+   if (argc > 2)
    {
-      argv_str[i] = argv[i + 2].getString();
-   }
-   bool result;
-   const char* methodRes = CInterface::CallMethod(object, funcName, argv_str, argc - 2, &result);
+      const char** argv_str = static_cast<const char**>(malloc((argc - 2) * sizeof(char*)));
```
jamesu: Do we not have a typed malloc function for this? Might be cleaner

Engine/source/persistence/taml/taml.cpp
```diff
@@ -644,18 +644,20 @@ ImplementEnumType(_TamlFormatMode,
...
+         if (!defaultConObject)
+            return;
          defaultObject = dynamic_cast<SimObject*>(defaultConObject);
       
-         // ***Really*** shouldn't happen
-         if (!defaultObject)
-            return;
       }
+      // ***Really*** shouldn't happen
+      if (!defaultConObject || !defaultObject)
```
jamesu, line 659: defaultObject will be NULL if the class is derived from NetEvent. defaultConObject will be NULL if an invalid class is specified.

Perhaps a safer approach here would be to make some sort of ConsoleObject::create_simobject_class function which only creates classes if they are derived from SimObject. Perhaps it could be expanded to only allow certain class types to be constructed from a taml file.

Engine/source/console/console.cpp
Azaezel: not at time of writing

Engine/source/persistence/taml/taml.cpp
Azaezel: in the interest of letting lukas complete https://github.com/TorqueGameEngines/Torque3D/pull/1004 in a timely manner, we'll toss this in with a note that in the context it's used for, the method is for dumping simobjects to and from taml files, so neither scenario occurs. though we will likely want to circle back for cleanliness's sake

PR #1020, opened 2023-05-11
FreeBSD ready

ohmtal:
Compiles on FreeBSD but libs:  pcre  and convexDecomp are altered to work.  

Fixed: I have a problem with the posix CPUInfo, because std::ptr_fun is obsolete. 

Added posix fix for onExit. (https://torque3d.org/forums/topic/7261-fix-for-onexit-not-called-on-linuxbsd-posix/)

### DIFF
--- a/Engine/lib/convexDecomp/NvSimpleTypes.h
+++ b/Engine/lib/convexDecomp/NvSimpleTypes.h
@@ -57,10 +57,15 @@ NvSimpleTypes.h : Defines basic data types for integers and floats.
 
 */
 
+
 #ifdef __APPLE__
    #include <sys/malloc.h>
 #else
+#if defined( __FreeBSD__)
+   #include <stdlib.h>
+#else 
    #include <malloc.h>
+#endif
 #endif
 #include <assert.h>
 
@@ -87,7 +92,7 @@ NvSimpleTypes.h : Defines basic data types for integers and floats.
 	typedef float				NxF32;
 	typedef double				NxF64;
 		
-#elif defined(LINUX)
+#elif LINUX
 	typedef long long			NxI64;
 	typedef signed int			NxI32;
 	typedef signed short		NxI16;
@@ -101,7 +106,7 @@ NvSimpleTypes.h : Defines basic data types for integers and floats.
 	typedef float				NxF32;
 	typedef double				NxF64;
 
-#elif defined(__APPLE__)
+#elif __APPLE__
 	typedef long long			NxI64;
 	typedef signed int			NxI32;
 	typedef signed short		NxI16;
@@ -115,7 +120,7 @@ NvSimpleTypes.h : Defines basic data types for integers and floats.
 	typedef float				NxF32;
 	typedef double				NxF64;
 
-#elif defined(__CELLOS_LV2__)
+#elif __CELLOS_LV2__
 	typedef long long			NxI64;
 	typedef signed int			NxI32;
 	typedef signed short		NxI16;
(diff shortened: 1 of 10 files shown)


Ragora:
I feel like there's probably a better way to do the trims I had in there (maybe not, though):

https://stackoverflow.com/questions/216823/how-to-trim-an-stdstring

```c++
// trim from start (in place)
static inline void ltrim(std::string &s) {
    s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](unsigned char ch) {
        return !std::isspace(ch);
    }));
}

// trim from end (in place)
static inline void rtrim(std::string &s) {
    s.erase(std::find_if(s.rbegin(), s.rend(), [](unsigned char ch) {
        return !std::isspace(ch);
    }).base(), s.end());
}
```

Torque might have a usable function as well, but I dunno what it would be offhand.

ohmtal:
> I feel like there's probably a better way to do the trims I had in there (maybe not, though):
> 
> https://stackoverflow.com/questions/216823/how-to-trim-an-stdstring
> 
> ```c++
> // trim from start (in place)
> static inline void ltrim(std::string &s) {
>     s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](unsigned char ch) {
>         return !std::isspace(ch);
>     }));
> }
> 
> // trim from end (in place)
> static inline void rtrim(std::string &s) {
>     s.erase(std::find_if(s.rbegin(), s.rend(), [](unsigned char ch) {
>         return !std::isspace(ch);
>     }).base(), s.end());
> }
> ```
> 
> Torque might have a usable function as well, but I dunno what it would be offhand.

I've added your changes.

Engine/lib/convexDecomp/NvSimpleTypes.h
```diff
@@ -87,7 +92,7 @@ NvSimpleTypes.h : Defines basic data types for integers and floats.
 	typedef float				NxF32;
 	typedef double				NxF64;
 		
-#elif defined(LINUX)
+#elif LINUX
```
dottools, line 95: What is going on here with removing `defined()` evaluations?

Engine/lib/convexDecomp/NvThreadConfig.cpp
```diff
@@ -129,26 +133,22 @@ void   tc_sleep(NxU32 ms)
...
-#elif defined( _WIN64 )
-   YieldProcessor( );
-#elif defined( __APPLE__ )
-   pthread_yield_np();
-#elif defined(__linux__)
-   #if defined(_POSIX_PRIORITY_SCHEDULING)
-      sched_yield();
+   #if defined(__linux__) || defined( __APPLE__ ) ||  defined( __FreeBSD__)
+
+      asm ( "pause" );
```
dottools, line 138: Pretty sure we've already previously had discussions about using inlined x86 assembly instruction, vs using the preferred way of involving the OS scheduler in yielding the task and for cross architecture compatibility (`pause` doesn't exist for ARM, etc..).

Engine/source/platformPOSIX/POSIXCPUInfo.cpp
```diff
@@ -20,7 +20,8 @@
 // IN THE SOFTWARE.
 //-----------------------------------------------------------------------------
 
-#ifndef __APPLE__
+//XXTH orig #ifndef __APPLE__
+#if not defined (__APPLE__) //and not defined (__FreeBSD__)
```
dottools, line 24: `#if not defined (__APPLE__)` is not a valid/standard preprocessor expression.

I believe you meant
`#if !defined (__APPLE__)`

ohmtal:
Thanks for the review dottools. I think it need some cleanup. Will close it for now.

PR #1027, opened 2023-05-27
update sdl to 2.26.5

Azaezel:


Azaezel:
compile fix encapsulated into #1032

PR #1062, opened 2023-07-22
Unit tests in same project/Separate Main

marauder2k7:
This is a test of the GTests when they are built slightly differently.


marauder2k7:
look to latest, rolled in with cmake overhaul

PR #1063, opened 2023-07-23
This implements a fix to an issue with the CICD that causes a segfault.

Areloch:
The fix adds a sanity check to the D3D and GL device enumeration function that exits early if there's no registered display from the OS(implying it's running in commandline). With newer versions of SDL, attempting to enumerate the GL device causes a segfault because the OS will return back a valid context, but the context isn't actually valid. So when tested against, it crashes. Avoiding enumerating the device when you're not in a position to render works around the issue.

TODO: If the machine is running in terminal-only mode, and thus has no valid render context, but still has a monitor plugged into the machine, it can sidestep the sanity check and still result in a segfault. Need a more robust check for that circumstance in a future fix.


lukaspj:
I wonder if this solves the issue I had in #405

PR #1064, opened 2023-07-24
Unit tests separated with CMake Overhaul pr

marauder2k7:
TEST PR: do not commit before cmake overhaul


marauder2k7:
closed for different tests

PR #1075, opened 2023-08-26
Bin optimization

jamesu:
Replaces the linked lists used in the SceneContainer bin storage system and the zoning system with contiguous vectors.

Note that to facilitate fast lookup of objects in bins and zones, a new SceneContainerBinRefList<> class is used. This manages a list of structs which point to items in a contiguous list of bin indexes. Objects reserve these structs via handles in order to manage their associated bin or zone indexes. The list will occasionally get re-compacted once a threshold of struct handles have been released.

Please also note that this code currently changes the enumeration order for bins. Normally in T3D objects end up getting inserted and removed in FILO order (due to objects being in a linked list), while this code ends up using FIFO order. Enumeration order might also be affected when objects are removed, as erase_fast is used. I suspect this likely won't be a problem though unless for example someone is using the journalling functionality and wants reproduction of events to be exactly the same.

Unit tests are also included, including some for the existing SceneContainer functionality.


jamesu:
NOTE: Need to repro the mine collision issue with this one before merging. See https://discord.com/channels/358091480004558848/783126640712613899/1146104819023630417

PR #1076, opened 2023-09-05
Tweaks some of the handling for timestamp stuff

Areloch:
From Nils' UI Update work. Tweaks some of the handling for timestamp stuff, and adds some utility console functions for getting/resetting said timestamp stuffs.


Engine/source/console/console.cpp
```diff
@@ -307,6 +307,9 @@ bool alwaysUseDebugOutput = true;
 bool useTimestamp = false;
 bool useRealTimestamp = false;
 
+static U32 initTime = Platform::getRealMilliseconds();
+U32 startTime = initTime;
```
dottools, line 311: should be `static U32 startTime = initTime;` to avoid name collision outside of source file.

PR #1086, opened 2023-09-13
TScript - Fix Type Bug

JeffProgrammer:
The compiler was found to have a bug where the variable type cannot ever change, causing issues.


lukaspj:
Maybe create a regression test for this?

PR #1095, opened 2023-09-22
Shader Blueprint class for a unified shader language

marauder2k7:
- ShaderBlueprint class
- Basic unified shader language parser that exports both hlsl and glsl files 
- Simple example shaderblueprint included.


lukaspj:
At a glance, two points:

1. A documentation of the spec for these blueprint files would be nice. It can be added as a Doxygen comment like [this](https://github.com/TorqueGameEngines/Torque3D/blob/development/Engine/source/console/consoleObject.h#L114-L176)
2. I don't think ".blueprint" is a good extension, it's too generic. I'd suggest either ".tsl" (Torque shader language), ".usl" (unified shader language), ".shader" or something similar.

PR #1104, opened 2023-10-03
SFXPlaylist asset groundwork

marauder2k7:
-Reverted SFXPlaylist since it is going to be made from an asset now instead. 
-Added extra options to soundAssets description.
-SFXPlaylist may need an onAdd function


marauder2k7:
Rebase messed up with this PR gonna create a new one from the working state

PR #1148, opened 2023-11-30
Stb image loading

marauder2k7:
Mark this for 4.2

NOTES: Screenshot needs updated to use stb, was originally able to tile for larger than normal res screenshots need to figure ot if this is possible in stb image or if we need to just do it in gbitmap.

blTerrainSystem lightmap needs to be changed to use a filename, or if we could figure out how to marry Stream and the stbimage loaders we could revert back from using a filepath and use stream instead.


Engine/source/gfx/screenshot.cpp
```diff
@@ -130,85 +130,85 @@ void ScreenShot::capture( GuiCanvas *canvas )
...
+
+   //      delete gb;
+   //   }
+
+   //   // Write the captured tile row into the PNG stream
+   //   pngWriter.append(outBuffer, outBuffer->getHeight()-overlapPixels.y);
+   //}
+
+   ////Close the PNG stream
+   //pngWriter.end();
```
dottools: Please learn to:
```cpp
#if 0
// code to disable goes here
#endif // 0
```

Instead of multi-line commenting out block of code to avoid these unnecessary large diffs.

marauder2k7: > Please learn to:
> 
> ```c++
> #if 0
> // code to disable goes here
> #endif // 0
> ```
> 
> Instead of multi-line commenting out block of code to avoid these unnecessary large diffs.

was talked over with az and jeff, this is just commented out for now to be changed when we figure out how to do this with stb image, essentially it is mostly going to be removed either way

Engine/source/gfx/bitmap/loaders/bitmapSTB.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+   if (ext.equal("bmp"))
+   {
+      if (stbi_write_bmp(path.getFullPath().c_str(), width, height, comp, bitmap->getWritableBits()))
+         return true;
+   }
+
+   if (ext.equal("jpg") || ext.equal("jpeg"))
+   {
+      if (stbi_write_jpg(path.getFullPath().c_str(), width, height, comp, bitmap->getWritableBits(), 90))
```
dottools: Quality level should be a configurable value not hard coded to 90.

Engine/source/gfx/screenshot.cpp
dottools: I understand, but for languages `C`, `C++`, and even `C#` multiline commenting out is just not acceptable for revision control when you can `#if 0` it out to disable code for the time being until it is actually revised or actually deleted.

marauder2k7: sorted

Engine/source/gfx/bitmap/loaders/bitmapSTB.cpp
marauder2k7: this one i was thinking of using the compressionLevel argument in the function, but people wold need to know that it is a percentage so needs to be properly documented

dottools: Ah ok. Here's an idea for documenting this function:
```cpp
/**
 * Write bitmap to an image file.
 *
 * @param[in]       path                        Destination image file path. File name extension determines image format.
 *                                              ".bmp" for Microsoft Bitmap (BMP).
 *                                              ".hdr" for High Dynamic Range (HDR).
 *                                              ".jpg" or ".jpeg" for Joint Photographic Experts Group (JPEG).
 *                                              ".png" for Portable Network Graphics (PNG).
 *                                              ".tga" for Truevision TGA (TARGA).
 *
 *
 * @param[in]       bitmap                      Source bitmap to encode image from.
 * @param           compressionLevel            Image format specific compression level.
 *                                              For JPEG sets the quality level percentage, range 0 to 100.
 *                                              Not used for other image formats.
 */
bool sWriteSTB(const Torque::Path& path, GBitmap* bitmap, U32 compressionLevel)
```

marauder2k7: nice thanks :) ill put that in, and ill add the documentation to gbitmap as well since it comes directly from that

marauder2k7: committed

Azaezel:
need to resolve an application close flaw, but otherwise seems good

marauder2k7:
> need to resolve an application close flaw, but otherwise seems good

no longer able to trip this error after updating the pr to latest

PR #1150, opened 2023-12-03
Class prototyping presenter

Azaezel:
inspector hooked class method prototyping interface


Azaezel:
we'll have to put this one off till we sort unnamed objects and why it's popping up when the particle editor is activated

Templates/BaseGame/game/data/Prototyping/UI/classPrototyping.tscript
```diff
@@ -0,0 +1,145 @@
...
+function classPrototyping::SetNamespaceUsed(%this, %nameSpaceUsed)
+{
+    ClassMethodListCtrl.deleteAllObjects();
+    if (%nameSpaceUsed $= %this.instanceName)
+    {
+        %this.tempClass = %this.inst;
+        %this.fillMethodlist(%this.inst);
+        return;
+    }
+    eval("classPrototyping.tempClass = new "@ %nameSpaceUsed @"(){};");
```
Azaezel: main blocker point for inclusion. this route requires instantiating a class of this inheritance level. which the lions share of interactive classes require a valid datablock for (meaning things like rigidshape and it'd derivatives would need pointing at a datablock of the relevant type with a valid collision mesh for). Ideally we'd simply look up the class structure itself, rather than creating an actual object. Requires more research.

Azaezel: resolved though the new <foo>NS methodsig and field variant methods

PR #1165, opened 2023-12-28
RigidShape resolveContacts Interpenetration fix

AtomicWalrus:
Convert impulse units to force to match units of the spring model. This fixes the classic "sinking object" issue with Vehicle and RigidShape objects resting on surfaces or otherwise interacting at low speeds.

I suggest a contactTol between 0.5-2 m/s with this fix in place (depending on the object size and expected velocities of interaction).

Additional notes from my extra-commented version:
// Contact model has 2 components:
            // 1: Spring separates inter-penetrated surfaces over time (dist < collisionTol). Calculated using modified k*dx spring model (k ~ zi).
            // 2: No additional inter-penetration is allowed (zero all momentum into surface) by modifying k (zi). Calculated by getZeroImpulse -> impulse -> converted to force (12/23 fix)
            // Unit analysis notes on the original model:
            //    -"zi" roughly acts as a spring constant, but with units of Force instead of the more common Force/Dist
            //    -"collisionTol - distance" has units of distance, but is treated as a unitless scalar in the final force calculation.

### DIFF
--- a/Engine/source/T3D/rigidShape.cpp
+++ b/Engine/source/T3D/rigidShape.cpp
@@ -1350,7 +1350,8 @@ bool RigidShape::resolveContacts(Rigid& ns,CollisionList& cList,F32 dt)
 
             // Penetration force. This is actually a spring which
             // will seperate the body from the collision surface.
-            F32 zi = 2 * mFabs(mRigid.getZeroImpulse(r,c.normal));
+            // Contact fix 12/23: Convert "zero impulse" to force to match units of spring model. F = I / dt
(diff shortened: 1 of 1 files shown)


AtomicWalrus:
Just for historical reference since the actual PR that merged this change links here, the two patches to the contact solver were:
-Unit conversion of impulse to force to match spring eq (getZeroImpulse / dt) 
-Apply force immediately after each contact in the list, instead of accumulating, so that velocity changes apply before additional contacts are tested (otherwise results will be wrong and you'll see bouncing)

PR #1208, opened 2024-02-06
Datachunker & FrameAllocator refactor

jamesu:
There was interest in incorporating this code to hopefully help resolve some memory-related issues on other platforms, so here it is! 

This refactors both FrameAllocator and DataChunker to use a unified aligned memory allocator (AlignedBufferAllocator). A new ThreeTieredChunker helper class has been also been added to standardize the allocation pattern used in DecalManager.
In addition FrameAllocator has been made thread_local so as long as it is set up correctly it can be safely used in worker threads.

Also adds unit tests for everything changed, something terribly missing from the original code which has probably been hacked around one too many times by now.

Copyright header modified as this is fresh code (no other authors apart from me at this stage). Suggestions for any improvements welcome.


jamesu:
Added fix for gcc & clang

PR #1211, opened 2024-02-09
Test assimp PR before full refactor

marauder2k7:
non destructive fixes for assimp loader
-this creates an ambient sequence if 1 does not exist.
-adds a bounds node if 1 does not exist.


Azaezel:
I'd be generally inclined to include this regardless of whether or not they get back to us on https://github.com/assimp/assimp/issues/5459 as a step forward. We will of course need to make a note about flipping off NLA Strips and All Actions when exporting FBX from Blender to prevent the resulting merged animation channels from causing erroneous applications of transform information with independently mobile submeshes. (Simplified example case included in the report their end. Amounts to an inability to animate non-armature based things like brick explosions without those two flags off)

marauder2k7:
> I'd be generally inclined to include this regardless of whether or not they get back to us on [assimp/assimp#5459](https://github.com/assimp/assimp/issues/5459) as a step forward. We will of course need to make a note about flipping off NLA Strips and All Actions when exporting FBX from Blender to prevent the resulting merged animation channels from causing erroneous applications of transform information with independently mobile submeshes. (Simplified example case included in the report their end. Amounts to an inability to animate non-armature based things like brick explosions without those two flags off)

sure thing, if you want to commit it go ahead, there will be a follow up refactor of this with further functionality, i will update this pr with main now and get it ready to be pushed.

PR #1216, opened 2024-02-14
Geometry shaders, drawRoundedRect

marauder2k7:
Added support for geometry shaders
drawRoundedRect function in gfxDrawUtil (to be replaced with geometry shaders)

DO NOT MERGE: this needs to go through extensive testing. and all gui draw calls need to be changed to use geometry shaders instead.

### DIFF
--- a/Engine/source/gfx/gfxDrawUtil.h
+++ b/Engine/source/gfx/gfxDrawUtil.h
@@ -189,12 +189,12 @@ protected:
    FontRenderBatcher* mFontRenderBatcher;
 
    // Expanded shaders
-   GFXShaderRef mRoundRectangleSahder;
-   GFXShaderConstBufferRef mRoundRectangleSahderConsts;
+   GFXShaderRef mRoundRectangleShader;
+   GFXShaderConstBufferRef mRoundRectangleShaderConsts;
    GFXShaderConstHandle* mModelViewSC;
    GFXShaderConstHandle* mRadiusSC;
-   GFXShaderConstHandle* mSizeSc;
-   GFXShaderConstHandle* mRectCenterSc;
+   GFXShaderConstHandle* mSizeSC;
+   GFXShaderConstHandle* mRectCenterSC;
 };
 
 #endif // _GFX_GFXDRAWER_H_
--- a/Engine/source/gfx/gfxDrawUtil.h
+++ b/Engine/source/gfx/gfxDrawUtil.h
@@ -46,6 +46,15 @@ public:
    ~GFXDrawUtil();
 
    //-----------------------------------------------------------------------------
+   // Draw Ellipse
+   //-----------------------------------------------------------------------------
+   void drawEllipse(const Point2I& center, const U32& radiusX, const U32& radiusY, const ColorI& color, const U32& segments);
+
+   void drawRoundedRect(const U32& cornerRadius, const RectI& rect, const ColorI& color, const U32& segments);
+
+   void drawRoundedRect(const U32& cornerRadius, const Point2I& upperLeft, const Point2I& lowerRight, const ColorI& color, const U32& segments);
+
+   //-----------------------------------------------------------------------------
    // Draw Rectangles
    //-----------------------------------------------------------------------------
    void drawRect( const Point2F &upperLeft, const Point2F &lowerRight, const ColorI &color );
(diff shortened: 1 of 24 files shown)


Engine/source/gfx/gfxDrawUtil.h
```diff
@@ -187,6 +187,14 @@ class GFXDrawUtil
...
    
    FontRenderBatcher* mFontRenderBatcher;
+
+   // Expanded shaders
+   GFXShaderRef mRoundRectangleSahder;
+   GFXShaderConstBufferRef mRoundRectangleSahderConsts;
+   GFXShaderConstHandle* mModelViewSC;
+   GFXShaderConstHandle* mRadiusSC;
+   GFXShaderConstHandle* mSizeSc;
+   GFXShaderConstHandle* mRectCenterSc;
```
dottools: Typos and inconsistencies here. `Sahder` instead of `Shader` and `SC` vs `Sc`.

marauder2k7: thanks will sort those out

dottools:
Not to kill the mood with the development here, but you are aware that geometry shaders are discouraged and a huge performance hit in the graphics pipeline, right? The geometry shader was designed to only be used in a specialize manner. Not for common draw operations. Hence why its been effectively replaced with [mesh shaders in OpenGL](https://developer.nvidia.com/blog/introduction-turing-mesh-shaders/), [DirectX 12](https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html) and [Vulkan](https://www.khronos.org/blog/mesh-shading-for-vulkan).

So far the work I've seen in this PR suggests that a lot of common basic draw calls to always go through geometry shaders and it will not end well.

References:
 * [Why does this geometry shader slow down my program so much?](https://gamedev.stackexchange.com/questions/48432/why-does-this-geometry-shader-slow-down-my-program-so-much)
* [So, what's the state of the geometry shader these days? If it's still needlessly slow, then what are the better alternatives for its various use-cases?](https://www.reddit.com/r/GraphicsProgramming/comments/h8h3pb/so_whats_the_state_of_the_geometry_shader_these/)
* [Friday Facts #251 - A Fistful of Frames](https://www.factorio.com/blog/post/fff-251)

marauder2k7:
> Not to kill the mood with the development here, but you are aware that geometry shaders are discouraged and a huge performance hit in the graphics pipeline, right? The geometry shader was designed to only be used in a specialize manner. Not for common draw operations. Hence why its been effectively replaced with [mesh shaders in OpenGL](https://developer.nvidia.com/blog/introduction-turing-mesh-shaders/), [DirectX 12](https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html) and [Vulkan](https://www.khronos.org/blog/mesh-shading-for-vulkan).
> 
> So far the work I've seen in this PR suggests that a lot of common basic draw calls to always go through geometry shaders and it will not end well.
> 
> References:
> 
> * [Why does this geometry shader slow down my program so much?](https://gamedev.stackexchange.com/questions/48432/why-does-this-geometry-shader-slow-down-my-program-so-much)
> * [So, what's the state of the geometry shader these days? If it's still needlessly slow, then what are the better alternatives for its various use-cases?](https://www.reddit.com/r/GraphicsProgramming/comments/h8h3pb/so_whats_the_state_of_the_geometry_shader_these/)
> * [Friday Facts #251 - A Fistful of Frames](https://www.factorio.com/blog/post/fff-251)

So the general idea is to use geometry shaders for things that would bog down the cpu for certain draw calls such as point sprites, bezier curves and a few other things. We are updating the gfxDrawutil to do more with pixel shaders such as the rounded rectangle bit but as for the geometry shader end they will only be used for specific cases.

one other good use for them is to draw simple shapes such as arrows for gizmos, exploding meshes, rendering normals etc these are done faster rather than looping through and defining lines on the cpu to then pass as a draw call, since each line would need to be its own draw call, whereas with a geometry shader the visualization of the normals of a mesh would be one draw call as shown here https://learnopengl.com/Advanced-OpenGL/Geometry-Shader

Other expanded uses are being considered but if its a slower workflow in geometry shaders, we will switch them over to compute shaders as well. 

Voxel cone ssgi is another use case for geometry shaders as it voxelizes the frustum geometry to create surfaces for rendering lighting and reflectance

Templates/BaseGame/game/core/rendering/shaders/fixedFunction/thickLineG.hlsl
```diff
@@ -0,0 +1,78 @@
...
+};
+
+struct PSConn
+{
+   float4 HPOS             : TORQUE_POSITION;
+   float4 color            : COLOR;
+};
+
+uniform float thickness;
+uniform float4x4 modelview;
```
marauder2k7, line 38: /

marauder2k7:
```suggestion
```

marauder2k7:
has been rolled into the GFX Shader Const refactor PR closing this

PR #1218, opened 2024-02-21
Loads an IES Photometric profile.

marauder2k7:
ADDED: Ability to add IES profile as the cookie texture slot in both point lights and spot lights

TODO:
Have the IES Profile also drive the settings for the lights. (future updates to lighting required)
Make it work in tandum with Cookie textures. :ballot_box_with_check:
IES profiles are to be another slot in the advanced light section. :ballot_box_with_check:


marauder2k7:
Some notes: This is ready to be rolled in, going to upload a few example ies profiles for now into prototyping. 

The way forward with this kind of setup is to use ies profile data to feed light properties before it reaches the shader such as light intensity, color etc. Best used with point lights at the moment as they have the full range required for an ies, using with a spotlight requires the spotlight outer cone to be set to 180 on most ies profiles. Again these settings can come from the ies profile itself so something to look out for in future.

PR #1229, opened 2024-03-15
macosx changes

marauder2k7:
changes required to get masox to run.

Some further changes are possibly required around resourceHolder in future.

The geometry shader works in macosx with gl version 3.3, though i do think some changes are required internally to account for a lack of OUT_col

### DIFF
--- a/Engine/source/core/resource.h
+++ b/Engine/source/core/resource.h
@@ -194,7 +194,7 @@ template<class T> class ResourceHolder : public ResourceHolderBase
 {
 public:
    ResourceHolder(T *t) : ResourceHolderBase(t) {}
-   virtual ~ResourceHolder() { delete ((T*)mRes); }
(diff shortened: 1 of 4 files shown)


Engine/source/core/resource.h
```diff
@@ -194,7 +194,7 @@ template<class T> class ResourceHolder : public ResourceHolderBase
 {
 public:
    ResourceHolder(T *t) : ResourceHolderBase(t) {}
-   virtual ~ResourceHolder() { delete ((T*)mRes); }
+   virtual ~ResourceHolder() { mRes = static_cast<T*>(mRes); SAFE_DELETE(mRes); }
```
jamesu: I would strongly suggest adding a unit test to make sure ~ResourceHolder actually calls the destructor correctly considering this is undefined behavior and could easily break without knowing.

PR #1230, opened 2024-03-17
TEST PR do not commit this

marauder2k7:
if we are in linux, define _CRT_USE_BUILTIN_OFFSET


marauder2k7:
some interesting tests run here, all data from classes can be put into structs and this kills the offsetof warning. more tests to be done to see if this affects performance and script side.

PR #1233, opened 2024-03-18
Unnecessary virtuals

marauder2k7:
This PR is to remove unnecessary virtuals from sub classes:

SceneObject derived classes done.

TODO:
GuiControl derived classes
GFX derived classes
GameObject derived classes

### DIFF
--- a/Engine/source/environment/waterObject.h
+++ b/Engine/source/environment/waterObject.h
@@ -63,7 +63,7 @@ class MatrixSet;
 
 struct WaterMatParams
 {
-   MaterialParameterHandle* mRippleMatSC;   
+   MaterialParameterHandle* mRippleMatSC;
    MaterialParameterHandle* mRippleDirSC;
    MaterialParameterHandle* mRippleTexScaleSC;
    MaterialParameterHandle* mRippleSpeedSC;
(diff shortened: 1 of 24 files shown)


Engine/source/environment/waterObject.h
```diff
@@ -156,33 +156,33 @@ class WaterObject : public SceneObject
...
 
    // SimObject
-   virtual bool onAdd();
-   virtual void onRemove();
-   virtual void inspectPostApply();
-   virtual bool processArguments(S32 argc, ConsoleValue *argv);
+   bool onAdd();
+   void onRemove();
+   void inspectPostApply();
+   bool processArguments(S32 argc, ConsoleValue *argv);
```
dottools, line 162: This is going to break `WaterPlane` class. Please pay attention as to what you're doing here.

All SimObject methods like this should always be virtual when the object might be inherited in the future.
Same goes for linage of `NetObject`, as I'm seeing the same for `packUpdate`/`unpackUpdate` methods. and other commonly inherited classes.

Do not trust Visual Studio's static analyzers without manually reviewing each case because it is often very wrong.

marauder2k7: i have tested all the classes i have changed by loading up an instance of them in the world editor and they work fine is there another test i should do?

marauder2k7:
close to remove unnecessary line changes

PR #1252, opened 2024-04-10
Templated math functions

marauder2k7:
change all the code to use templated math functions. 

All calls to these math functions now enforce using the correct types (can no longer compare int to float for example)


marauder2k7:
out of date needs to be re-done

PR #1263, opened 2024-04-25
Add cstdint to ies_loader

SKBotNL:
On a Ubuntu install with gcc 11.4.0, Torque3D builds fine without any modifications. On my Void Linux install however, which uses gcc 13.2.0, `Engine/source/gfx/bitmap/loaders/ies/ies_loader.cpp` and `Engine/source/gfx/bitmap/loaders/ies/ies_loader.h` fail to build.
Adding `#include <cstdint>` to these files fixes the issue, I have also tested this modification on Ubuntu (with gcc 11) and it builds just fine.

### DIFF
--- a/Engine/source/gfx/bitmap/loaders/ies/ies_loader.cpp
+++ b/Engine/source/gfx/bitmap/loaders/ies/ies_loader.cpp
@@ -37,7 +37,6 @@
 #include "ies_loader.h"
 #include <assert.h>
 #include <algorithm>
-#include <cstdint>
 #include <functional>
 
 #include "math/mMathFn.h"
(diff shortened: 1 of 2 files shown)


marauder2k7:
good catch. Do the IES profiles still work with this change in? Havent actually tested to see how they look on linux.

SKBotNL:
> good catch. Do the IES profiles still work with this change in? Havent actually tested to see how they look on linux.

I'm pretty new to this engine (and game development as a whole), how do I test that?

SKBotNL:
After adding lighting (spotlight) using the world editor, and assigning a .IES file from Phillips. It seems like the actual light just... disappears. Should I be using a specific .IES file?

Azaezel:
seems sane. though I'll note its redundant to add the include in both the .h and the .cpp that includes the .h

SKBotNL:
> seems sane. though I'll note its redundant to add the include in both the .h and the .cpp that includes the .h

So only the include in the .h is needed?

Azaezel:
> > seems sane. though I'll note its redundant to add the include in both the .h and the .cpp that includes the .h
> 
> So only the include in the .h is needed?

yes

SKBotNL:
> > > seems sane. though I'll note its redundant to add the include in both the .h and the .cpp that includes the .h
> > 
> > 
> > So only the include in the .h is needed?
> 
> yes

Fixed

PR #1276, opened 2024-05-12
Remove nvConvexDecomp and replace it with v-hacd

marauder2k7:
Updated the lib used for convex hull generation.

note: needs tested on other platforms.


dottools:
Why are there unrelated changes in this PR? Such as `ThreadPool` to `TorqueThreadPool`?

marauder2k7:
> Why are there unrelated changes in this PR? Such as `ThreadPool` to `TorqueThreadPool`?

naming conflict with the resource as it also has a class called threadpool

marauder2k7:
> Why are there unrelated changes in this PR? Such as `ThreadPool` to `TorqueThreadPool`?

turns out the ThreadPool class in the resource is nested in a namespace so we should be fine......... reverted rename back

PR #1289, opened 2024-06-16
if statements

marauder2k7:
Changed:
if check on vals now return true if the value has a string value %val = "test me" if(%val) will now return true since %val is not null

Script side:
string checks for "true" and "false" will now be parsed as integer values of 1 and 0.

TEST VIGOUROUSLY

### DIFF
--- a/Engine/source/console/console.h
+++ b/Engine/source/console/console.h
@@ -246,10 +246,10 @@ public:
          return s == StringTable->EmptyString() ? 0 : dAtoi(s);
       if (type == ConsoleValueType::cvString)
       {
-         if (dStrcmp(s, "false") == 0) {
+         if (dStricmp(s, "false") == 0) {
             return 0;
          }
-         else if (dStrcmp(s, "true") == 0) {
+         else if (dStricmp(s, "true") == 0) {
             return 1;
          }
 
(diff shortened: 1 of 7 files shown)


dottools:
Sounds like a really bad idea to implicitly convert strings `true` to 1 and `false` to 0. If that conversion is wanted then that should be done in a utility function, not performed on all string handling.

Because there are all kinds of situations where you are using `true` and `false` instead of integers for good reason, such as JSON handling.

marauder2k7:
> Sounds like a really bad idea to implicitly convert strings `true` to 1 and `false` to 0. If that conversion is wanted then that should be done in a utility function, not performed on all string handling.
> 
> Because there are all kinds of situations where you are using `true` and `false` instead of integers for good reason, such as JSON handling.

this is implicitly just for tscript file handling, it shouldnt effect json parsing, we are testing everything out now before this gets merged.

dottools:
What if I am parsing JSON within TorqueScript? Do not forget that TorqueScript is a general purpose scripting language and it is being used as such. You have to be careful with overly implicit operations that will break code that otherwise has been working fine. JSON was just an example, however `true` and `false` have been used in custom GUI controls I've written in the past that are explicitly used as flags for specific control behavior settings.

marauder2k7:
> What if I am parsing JSON within TorqueScript? Do not forget that TorqueScript is a general purpose scripting language and it is being used as such. You have to be careful with overly implicit operations that will break code that otherwise has been working fine. JSON was just an example, however `true` and `false` have been used in custom GUI controls I've written in the past that are explicitly used as flags for specific control behavior settings.

and this implicitly sets out to fix the issue with setting a var as true from source since it gets set as a string which if tested:
if(%var) was always returning false
we are working more on this to see if there is a possibility for a different fix but for now its the best we got.

PR #1294, opened 2024-07-22
various warnings

marauder2k7:
warnings cleanup

just fixing a few warnings that have been bugging for a while

### DIFF
--- a/Engine/source/core/util/uuid.cpp
+++ b/Engine/source/core/util/uuid.cpp
@@ -146,7 +146,14 @@ static void create_uuid_state(uuid_state *st)
  */
 static void format_token(char *target, const xuuid_t *u)
 {
-   snprintf(target, strlen(target), "%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x",
+   // first loop, figure out the size needed.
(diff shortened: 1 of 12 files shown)


Engine/source/core/util/uuid.cpp
```diff
@@ -146,11 +146,11 @@ static void create_uuid_state(uuid_state *st)
  */
 static void format_token(char *target, const xuuid_t *u)
 {
-  sprintf(target, "%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x",
-	  u->time_low, u->time_mid, u->time_hi_and_version,
-	  u->clock_seq_hi_and_reserved, u->clock_seq_low,
-	  u->node[0], u->node[1], u->node[2],
-	  u->node[3], u->node[4], u->node[5]);
+   snprintf(target, sizeof(target), "%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x",
```
dottools: `sizeof(target)` is getting size of the pointer, not buffer capacity. Buffer capacity is not being passed to the function.

marauder2k7: fixed

PR #1297, opened 2024-07-28
Template Matrix class

marauder2k7:
Adds a template class for matrices.

NOTE: Templated class has not replaced MatrixF yet.


dottools:
The thing that stands out to me immediately that needs work is the inconsistent brace coding format style.
Should match the cleaner brace coding format style that is already present in the `mMatrix.h` and `mMatrix.cpp` files.
Example, from awful bracing style:
```cpp
template<typename DATA_TYPE, U32 rows, U32 cols>
inline bool Matrix<DATA_TYPE, rows, cols>::isIdentity() const {
   for (U32 i = 0; i < rows; i++) {
      for (U32 j = 0; j < cols; j++) {
         if (j == i) {
            if((*this)(i, j) != static_cast<DATA_TYPE>(1)) {
               return false;
            }
         }
         else {
            if((*this)(i, j) != static_cast<DATA_TYPE>(0)) {
               return false;
            }
         }
      }
   }
   
   return true;
}
```

To proper bracing style that promotes readability:
```cpp
template<typename DATA_TYPE, U32 rows, U32 cols>
inline bool Matrix<DATA_TYPE, rows, cols>::isIdentity() const
{
   for (U32 i = 0; i < rows; i++)
   {
      for (U32 j = 0; j < cols; j++)
      {
         if (j == i)
         {
            if((*this)(i, j) != static_cast<DATA_TYPE>(1))
            {
               return false;
            }
         }
         else
         {
            if((*this)(i, j) != static_cast<DATA_TYPE>(0))
            {
               return false;
            }
         }
      }
   }

   return true;
}
```

There's several rather large inline functions in `mMatrix.h` that should be moved as normal class functions to `mMatrix.cpp`. Keep in mind that even though modern processors are fast, they're still constrained to the same instruction cache line sizes that existed decades prior and thus heavily duplicated machine generated code (inlined function) is slower and less efficient than multiple branched function calls to the same code memory location (normal function).

I'm also very concerned with all the `if ()` checks for `row` and `col`, and having to use nested loops. This templated matrix class is going to perform horribly compared to the per hand made matrix sized functions. The compiler won't be able to do opportunistic SIMD instruction generation with this class at all.

marauder2k7:
> The thing that stands out to me immediately that needs work is the inconsistent brace coding format style. Should match the cleaner brace coding format style that is already present in the `mMatrix.h` and `mMatrix.cpp` files. Example, from awful bracing style:
> 
> ```c++
> template<typename DATA_TYPE, U32 rows, U32 cols>
> inline bool Matrix<DATA_TYPE, rows, cols>::isIdentity() const {
>    for (U32 i = 0; i < rows; i++) {
>       for (U32 j = 0; j < cols; j++) {
>          if (j == i) {
>             if((*this)(i, j) != static_cast<DATA_TYPE>(1)) {
>                return false;
>             }
>          }
>          else {
>             if((*this)(i, j) != static_cast<DATA_TYPE>(0)) {
>                return false;
>             }
>          }
>       }
>    }
>    
>    return true;
> }
> ```
> 
> To proper bracing style that promotes readability:
> 
> ```c++
> template<typename DATA_TYPE, U32 rows, U32 cols>
> inline bool Matrix<DATA_TYPE, rows, cols>::isIdentity() const
> {
>    for (U32 i = 0; i < rows; i++)
>    {
>       for (U32 j = 0; j < cols; j++)
>       {
>          if (j == i)
>          {
>             if((*this)(i, j) != static_cast<DATA_TYPE>(1))
>             {
>                return false;
>             }
>          }
>          else
>          {
>             if((*this)(i, j) != static_cast<DATA_TYPE>(0))
>             {
>                return false;
>             }
>          }
>       }
>    }
> 
>    return true;
> }
> ```
> 
> There's several rather large inline functions in `mMatrix.h` that should be moved as normal class functions to `mMatrix.cpp`. Keep in mind that even though modern processors are fast, they're still constrained to the same instruction cache line sizes that existed decades prior and thus heavily duplicated machine generated code (inlined function) is slower and less efficient than multiple branched function calls to the same code memory location (normal function).
> 
> I'm also very concerned with all the `if ()` checks for `row` and `col`, and having to use nested loops. This templated matrix class is going to perform horribly compared to the per hand made matrix sized functions. The compiler won't be able to do opportunistic SIMD instruction generation with this class at all.

templated class, since c++ 11 and up all functions need to be inline otherwise you get unresolved externals errors. 

Due to the limited tests ive already done it is proving to be more performant than matrixf and the math_c functions

The reason for the if() checks on rows and cols is because this class is designed to be used with any size matrix you want, there will be implementations of 3x3 3x2 4x3 3x4 that we know of so far, for example the instancing stuff uses alot of 4x3 matrices, atm we have to take these in as 4x4, doesnt sound like much but theres over 70 of them in an array for the instancing shader

marauder2k7:
This PR is now fully ready for reviews and tests

to use the templated class add #define USE_TEMPLATE_MATRIX to the torqueConfig.h file

dottools:
Should add `#cmakedefine USE_TEMPLATE_MATRIX` with a comment in `Tools/CMake/torqueConfig.h.in` so that it can be a CMake build option.

marauder2k7:
> Should add `#cmakedefine USE_TEMPLATE_MATRIX` with a comment in `Tools/CMake/torqueConfig.h.in` so that it can be a CMake build option.

i was going to completely remove matrixF if this passed the review phase 

Added the cmake option, but it should be advices, if this gets committed into main while still in its review phase due to the number of classes effected by it and so on, its best to remove that cmake option

PR #1300, opened 2024-08-03
Macosx cmake cleanup and copy fix

marauder2k7:
Now the required shared libraries get copied into the sandbox.
OpenAL and SDL no longer complain about multiple RPATH
simplified macos configuration file, this should open the door also for other apple platform builds to be much simpler to implement eg ios


Engine/lib/CMakeLists.txt
```diff
@@ -199,8 +199,6 @@ add_subdirectory(glad ${TORQUE_LIB_TARG_DIRECTORY}/glad EXCLUDE_FROM_ALL)
 if(TORQUE_SFX_OPENAL)
     advanced_option(TORQUE_OGGVORBIS "Enable OGG Vorbis" ON)
     advanced_option(ALSOFT_EAX "Enable legacy EAX extensions" ${WIN32})
-    advanced_option(ALSOFT_INSTALL_EXAMPLES "Install example programs (alplay, alstream, ...)" OFF)
```
marauder2k7, line 202: these removed due to duplication

Tools/CMake/torqueMacOSconfigs.cmake
```diff
@@ -154,12 +54,14 @@ set(CMAKE_USE_WIN32_THREADS_INIT 0)
...
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 
+# Enable @rpath support
 set(CMAKE_MACOSX_RPATH 1)
-set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
 
-# Debug configuration for quicker debug builds
-set(CMAKE_XCODE_ATTRIBUTE_LINKER_DISPLAYS_MANGLED_NAMES[variant=Debug] YES)
-set(CMAKE_XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH[variant=Debug] YES)
-set(CMAKE_XCODE_ATTRIBUTE_ENABLE_TESTABILITY[variant=Debug] YES)
```
marauder2k7, line 163: these lines removed as they were being applied to all libraries. We should only change the settings like this on our own bundle build.

PR #1301, opened 2024-08-04
Templated point math type

marauder2k7:
Templated class for points.

More unit tests to be filled out.


marauder2k7:
Going to rethink this, point templated implementation is slightly more complicated due to a few things such as dimension access and how the class is declared throughout the codebase

PR #1313, opened 2024-09-12
World Editor GUI adjustments & corrections

Eikelenboom:
Some **World Editor GUI** changes:
1. Fixed the empty space (horizontal) between the world editor toolbars
2. Corrected the positions of the sidepanels _(which were off due to the introduction of the scene panels)_
3. Adjusted the window title font **aesthetics** _(because Arial has a very large width & height at size 18)_
4. Fixed the undefined background of the scene tab window
5. Correct Color Picker GUI text & text edit controls


Eikelenboom:
The change to the mission file got snuck in, obvious this should not be approved [here: ](https://github.com/TorqueGameEngines/Torque3D/pull/1313/files#diff-810378e43da896587e7137b05101599ff6779bb97961a92c720812e20629a938)

PR #1324, opened 2024-09-25
Add theme font options to editor settings & "InspectorTypeButtonProfile"  profile

Eikelenboom:
The text in the editor is displayed in the Arial font by default. This commit will add 2 x 2 extra options so it may be changed to the user preferences.

The theme is at its best with the Open Sans font so that text fields aren't cut off, because the width of the Arial characters is generally large. However, Arial remains to be default, also the fallback.

Console/engine function is added to be able to read environment variables in script; In this case to be able to find the fonts installation path.

-----------------

In _InspectorTypes.cpp_ there are referrals to a non-existing profile. That's why are the buttons in the inspector really hard to read and can't be changed in the prefs/settings.

Adding another profile "InspectorTypeButtonProfile" based on the default "ToolsGuiButtonProfile"

### DIFF
--- a/Engine/source/console/consoleFunctions.cpp
+++ b/Engine/source/console/consoleFunctions.cpp
@@ -2803,6 +2803,21 @@ DefineEngineFunction(getTimestamp, const char*, (), ,
    return returnBuffer;
 }
 
+//-----------------------------------------------------------------------------
+
+// Define the console function to get environment variable value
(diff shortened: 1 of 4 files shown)


Eikelenboom:
**Note:** These commits will update the GUI profiles with a check if the objects exist or not; at times the console got spammed with warnings (for example updating the theme in preferences).

I'm finished with this pull request. Cheers, Nils

PR #1333, opened 2024-11-27
SDF Font generation

marauder2k7:
needs work but the basic groundwork is there


marauder2k7:
I have a better method for doing this, the branch is out of date since the new changes and would make more sense to just use this branch as a reference.  Will have a new PR up for doing this.

PR #1343, opened 2024-12-27
Image Asset Refactor - For 4.2

marauder2k7:
All classes using image assets now go throught the image asset for the resource, before each class had a GFXTexHandle attached to them through the macros.

TypeImageAssetPtrRefactor replaces TypeImageAssetPtr and TypeImageAssetId for the script interfacing.


marauder2k7:
Sound Asset refactor will be its own PR keeping this one to just image assets

PR #1372, opened 2025-01-28
ShaderGen refactor 

marauder2k7:
allows passing a struct to a shaderGen feature. 

Usage Example:

```
// register feature with create function
FEATUREMGR->registerFeature(SNF_TextureFeature, new TextureFeatureHLSL, TextureFeatureHLSL::createFunction);

// create a struct for the feature
struct TextureFeatureParams
{
   String samplerName;
   String uvName;
   String uvType;

   TextureFeatureParams()
   {
      uvName = "texCoord";
   }
};

class TextureFeatureHLSL : public ShaderFeatureHLSL
{
private:
   /// Parameters that this feature can use to change the shadergen output.
   TextureFeatureParams* params;
public:
   /// default constructor
   NodeTextureFeatureHLSL()
   {
      params = new NodeTextureFeatureParams();
   }
   /// Constructor that takes params as an argument
   TextureFeatureHLSL(TextureFeatureParams* inParams)
   {
      params = inParams;
   }
  // create a static function on the feature class
  static ShaderFeature* createFunction(void* args)
  {
     TextureFeatureParams* params = static_cast<TextureFeatureParams*>(args);
      return new TextureFeatureHLSL(params);
  } 

   String getName() override
   {
      return "Texture Sampler";
   }
}
```


marauder2k7:
Re-Creating this PR without the unified wind deformation. Also rebasing.

PR #1422, opened 2025-03-24
Gametsctrl mouserefactor

Olathuss:
Refactored mouse callbacks in GameTSCtrl to use DECLARE_CALLBACK macros for consistency. Also added other mouse callbacks for completion, as well as a PlatformTimer to ensure mouse move/drag callbacks are not called every tick.


Olathuss:
Forgot to actually add the PlatformTimer changes, closed to add the changes and create a new pull request.

PR #1429, opened 2025-03-31
test adding new rule

marauder2k7:
add rule to evaluate -->foo[bar] to -->foobar


marauder2k7:
this wasnt a complete failure....... just partially one, need to add rules and specific sets that will do all the use cases, starting a new pr with the proposed language changes

PR #1494, opened 2025-06-06
Fixed major memory leak in GFXDevice::createStateBlock()

Nick-IronTower:
**Problem 1**

Due to the way the hashing function for GFXStateBlockDesc is implemented:
```
U32 GFXStateBlockDesc::getHashValue() const
{   
    return CRC::calculateCRC(this, sizeof(GFXStateBlockDesc));
}
```
...two seemingly identical GFXStateBlockDesc objects can produce different hashes. The reason behind this is memory padding. Extra bytes, added to the structure size by the compiler, contain undefined garbage data for each GFXStateBlockDesc instance created. Over time, GFXDevice::createStateBlock fills its mCurrentStateBlocks with an infinite number of GFXStateBlockDesc instances, each containing an identical set of useful data.

**Solution**

GFXStateBlockDesc memory packaging with no padding.

**Problem 2**

There is a similar issue with GFXSamplerStateDesc due to the implementation of operator==. Memory comparison doesn't provide a consistent and adequate result for the same reason.

**Solution**

GFXSamplerStateDesc memory packaging with no padding.

**Problem 3**

In GFXDevice::createStateBlock(), the following check doesn't work as intended:
`if (mCurrentStateBlocks[hashValue])`
...since Map's operator[] performs insertion.

**Solution**

Use Map::find() for the check.

**History**

This was a major issue for us after the release of The Age of Decadence, affecting a significant number of players. It took me some sleepless nights and looping through several memory analysis tools to catch it. Unfortunately, I had very little experience with GitHub at the moment, since we were using SVN internally, so I posted about this on GarageGames forums. I don't remember getting any replies, and the problem seemingly persists in this repository.
I'm not sure if it wasn't remedied in some other way, though, since I haven't followed T3D development for some time, so please kindly take a look.

### DIFF
--- a/Engine/source/gfx/gfxDevice.cpp
+++ b/Engine/source/gfx/gfxDevice.cpp
@@ -292,8 +292,9 @@ GFXStateBlockRef GFXDevice::createStateBlock(const GFXStateBlockDesc& desc)
    PROFILE_SCOPE( GFXDevice_CreateStateBlock );
 
    U32 hashValue = desc.getHashValue();
-   if (mCurrentStateBlocks[hashValue])
-      return mCurrentStateBlocks[hashValue];
+   auto it = mCurrentStateBlocks.find(hashValue);
+   if (it != mCurrentStateBlocks.end())
+      return it->value;
 
    GFXStateBlockRef result = createStateBlockInternal(desc);
    result->registerResourceWithDevice(this);   
--- a/Engine/source/gfx/gfxStateBlock.h
+++ b/Engine/source/gfx/gfxStateBlock.h
@@ -36,7 +36,7 @@
 #include "core/color.h"
 #endif
 
-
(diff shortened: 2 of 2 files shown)


Azaezel:
As a reminder note for those following along since we ran by the point fairly fast in the last memleak pass: 
For A|B testing, one can now compile a relwdebug aka OPTIMIZEDDEBUG binary compiled with TORQUE_DISABLE_MEMORY_MANAGER OFF via cmake or remmed in their torqueConfig.h
This will generate a memlog file listing uncleaned allocations with testing beginning at runtime and ending at ap close. it also introduced a LeakTrace(true/false) script command where true dumps the result to date and starts tracking mem allocations anew and false stops and dumps. Either route will list off a [count][totalmemleaked] + stacktrace.

PR #1511, opened 2025-06-29
Vulkan and Compute Support

marauder2k7:
Added the paths for compute shaders
Added the logic to allow unordered access views through texture objects Added the logic to bind these to ComputeShaders if they exist Added dispatch compute function to use the compute shader Use mLastComputeShader to control the flow of gfx dispatch. if we use compute shader, everything goes to it as compute should always be a separate stage.

TODO: OpenGL implementation. (requires opengl 4.3) 

Began the implementation of Vulkan to the graphics api.


marauder2k7:
Changing this PR to be for both the Vulkan implementation and the implementation of the compute pipeline.

This is so when we support compute on windows and linux we also support it on Mac

marauder2k7:
This PR dropped back in the history and requires a redo. If vcpkg is implemented then this will become a lot simpler so close for now as a history to check up on when doing the final implementation

PR #1576, opened 2025-10-14
ImageAsset caching images and add frames logic

marauder2k7:
Adds ability to either get the frame as a full texture, or get the frame that describes it 
Allows setting a number of uniform cells for an image.


marauder2k7:
this is all already in

PR #1579, opened 2025-10-18
VCPKG And Various Fixes

marauder2k7:
Added VCPKG to run the build on each platform. 

VCPKG finds dependencies and sets up the environment for our own dependencies to build properly.

Other Fixes: 
SFX System now integrates providers in a similar way the GFX system integrates adapters SFXDevice will setup automatically before the soundAssets are loaded meaning the buffers that need to be loaded should actually succeed.

MacOSX had an issue on teardown where it would try and free objects that didnt exist. This may have been causing issues on other platforms as well issue was in GuiPopUpCtrlEx

OpenAL Loading now uses system dll's as openal-soft is compiled using the al_router. 

Templates can now be installed without copying their .in files to the output directory

===================================================

LibSndFile operates with an LGPL license and needs to be linked dynamically
Assimp now has its dependencies compiled through vcpkg 
Faster build times since we are not compiling all the dependencies over and over again, vcpkg compiles them once.
Universal build now works on Mac


Azaezel:
to be very, very clear: while the idea has a great deal of merit, and we look forward to the kinks getting knocked off, there's enough vetting that needs to go into this one it's not going to make the 4.1 cutoff.

marauder2k7:
an updated pr for this exists.

PR #1582, opened 2025-10-24
Various Fixes and LIBSNDFILE precompiled

marauder2k7:


dottools:
Please do not include binaries into the repository. It caused issues in the past and is never a good idea. Have to make sure it all builds successfully before merging into the repository.

Areloch:
> Please do not include binaries into the repository. It caused issues in the past and is never a good idea. Have to make sure it all builds successfully before merging into the repository.

Out of curiosity, what were the prior issues examples?

This case came up because it turns out libSndFile is actually LGPL and so it requires being dynamically linked, but also the guy that wrote the build process mandated python as part of it for building it as a dynamic linkage. So it's a bit of a circumstantial mess.

dottools:
I recall issues using OpenAL when its binary was provided instead of loading it dynamically from the system for both Windows and Linux, although that might be back in the TGE days.

For Linux you should be loading it from the Linux distribution (provided by `libsndfile1` package on Debian based distros) due to system library dependencies. Biggest issue with prebuilt libraries and Linux has always been linking against `libc`, and now vs `musl libc`, that varies between Linux distribution variants and versions.

Regarding building complications from using Python. Is it really a big deal? Looks like it only requires the main Python package and nothing more.

Areloch:
> I recall issues using OpenAL when its binary was provided instead of loading it dynamically from the system for both Windows and Linux, although that might be back in the TGE days.
> 
> For Linux you should be loading it from the Linux distribution (provided by `libsndfile1` package on Debian based distros) due to system library dependencies. Biggest issue with prebuilt libraries and Linux has always been linking against `libc`, and now vs `musl libc`, that varies between Linux distribution variants and versions.
> 
> Regarding building complications from using Python. Is it really a big deal? Looks like it only requires the main Python package and nothing more.

I guess it mostly just an issue of 'seems silly to require grabbing python to run a handful of scripts to prep a library".

Areloch:
> I recall issues using OpenAL when its binary was provided instead of loading it dynamically from the system for both Windows and Linux, although that might be back in the TGE days.
> 
> For Linux you should be loading it from the Linux distribution (provided by `libsndfile1` package on Debian based distros) due to system library dependencies. Biggest issue with prebuilt libraries and Linux has always been linking against `libc`, and now vs `musl libc`, that varies between Linux distribution variants and versions.
> 
> Regarding building complications from using Python. Is it really a big deal? Looks like it only requires the main Python package and nothing more.

So, in terms of minimizing the speedbumps of people starting to getting the engine built: figuring that the python req is fine, what's your opinion on including a prebuilt of the python exec for running versus just telling the end user "just install python"?

The former feels conceptually similar to how we've got exe's bundled in for building bison and the like for the TS opcode stuff, but figured I'd get your opinion on if that strikes you as worth juggling or not vs trying to streamline the 'get the engine and go' experience.

Azaezel:
so 3 potential routes we're discussing:
1) make the end user install python in the os
2) an adapted cmake finder that points at a relative location of prebuilt binaries 
3) including python3 source and building that as a stage prior to libsnfile execution that *then* runs the finder

lukaspj:
“End user” is that game developers or gamers?

i have an example of getting hold of of Python via CMake if the developer installs the Python sdk

lukaspj:
Essentially this:
https://github.com/lukaspj/Torque3D/pull/4/files?diff=unified#diff-fd5a90d7dd4f09975a5ce323f7a5985ee5a8ffbf4baf000b6aec6319bdd180b2

Dunno if it helps any, not sure I understood what you are trying to solve 😅

Azaezel:
> “End user” is that game developers or gamers?
> 
> i have an example of getting hold of of Python via CMake if the developer installs the Python sdk

developers. yes. the library in question requires execution of python scripts to configure, or you can't even complete the cmake stage, let alone compile. the question is best practices for avoiding hassle on the part of new users just getting set up

marauder2k7:
> “End user” is that game developers or gamers?
> 
> i have an example of getting hold of of Python via CMake if the developer installs the Python sdk

End User would be the developer so they can build libsndfile as a dynamic library.

LibSndfile has some convoluted python script that builds symbols if you want to build it. Which since it is LGPL licensed we absolutely must do and must force.

advnau:
It is definitely key to getting this built and dynamically linked purely for licensing reasons. 

Is the python really a dependency still if vcpkg grabs it from the vcpkg port? Also since vcpkg can apply patches, maybe one of those could be generated to bypass python?

Id prefer not having a requirement on python, because at least on windows you have the microsoft store having app execution aliases for python/python3. Could serve as a headache for people if they havnt removed them.

marauder2k7:
> It is definitely key to getting this built and dynamically linked purely for licensing reasons.
> 
> Is the python really a dependency still if vcpkg grabs it from the vcpkg port? Also since vcpkg can apply patches, maybe one of those could be generated to bypass python?
> 
> Id prefer not having a requirement on python, because at least on windows you have the microsoft store having app execution aliases for python/python3. Could serve as a headache for people if they havnt removed them.

it is just vcpkg goes and gets a local python3 exe on windows and does it for you. 
If on linux you have to install python3 and pip install jinja2 for our dependencies. 
mac it also happens automatically for you

PR #1627, opened 2025-12-20
trip project name fill in popup before downloads

Azaezel:


Azaezel:
https://github.com/TorqueGameEngines/Torque3D/pull/1628 covers this too

PR #1642, opened 2025-12-28
ConvexFeature: Fix collision list object references

vitawrap:
Turns out ConvexFeature::collide assumes the second round of testVertex calls will only add one entry to the collision list at a time and was erroneously only changing the last object and material reference.

This annoying bug has been here since TGE as far as I can tell


Azaezel:
checks out with a good 75 rigidshapes vs a dense forest. much obliged.

Issue #1663, opened 2026-01-21
If client ControlObject is a camera set that to client.camera as well


marauder2k7:
From: Steve_Yorkshire https://discord.com/channels/358091480004558848/783127087820439582/1463578811768639603

If you spawn in as a camera, the controlObject is known as client.player ... but when drag/dropping an Ai/Player datablock from assetBrowser, it looks for localClientConnection.camera.getPosition() ... but client.camera has never been set. 

<img width="620" height="124" alt="Image" src="https://github.com/user-attachments/assets/9fea2f0d-9c16-4074-a89b-0c725dd049d6" /> 

Suggest fix 

Suggested fix; set %client.camera as the initial camera spawning in connectionToClient.tscript at onSetSpawnPointComplete

Result of fix.
<img width="1076" height="387" alt="Image" src="https://github.com/user-attachments/assets/f83a0506-2999-4b06-a109-03ef3e04badc" />

Version: Devhead #1660

PR #1664, opened 2026-01-21
address https://github.com/TorqueGameEngines/Torque3D/issues/1663

Azaezel:
by ensureing if the players is a camrea, then %conection.camera knows it's the same thing


marauder2k7:
resolves #1663

PR #1668, opened 2026-02-17
Various tweaks and fix for font cache

marauder2k7:
added fix for font cache (stb required size to be read switch back to libpng knocks the stream off)

Added extra control over the rendering of guibtimaps If guibitmaps are a child of buttons the profile colors now affect how to the button is rendered.

### DIFF
--- a/Engine/source/gfx/gFont.cpp
+++ b/Engine/source/gfx/gFont.cpp
@@ -705,10 +705,8 @@ bool GFont::read(Stream& io_rStream)
            delete bmp;
            return false;
        }*/
-       U32 len;
-       io_rStream.read(&len);
 
-       if (!bmp->readBitmapStream("png", io_rStream, len))
+       if (!bmp->readBitmapStream("png", io_rStream, U32_MAX))
        {
           delete bmp;
           return false;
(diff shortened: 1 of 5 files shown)


Engine/source/gfx/gFont.cpp
```diff
@@ -705,10 +705,8 @@ bool GFont::read(Stream& io_rStream)
            delete bmp;
            return false;
        }*/
-       U32 len;
-       io_rStream.read(&len);
 
-       if (!bmp->readBitmapStream("png", io_rStream, len))
+       if (!bmp->readBitmapStream("png", io_rStream, U32_MAX))
```
dottools, line 709: Whats the point of the length argument if you're going to lie to it by passing stream data length as nearly 4 GBytes?

marauder2k7: > Whats the point of the length argument if you're going to lie to it by passing stream data length as nearly 4 GBytes?

stb_reader required the length for bitmap streams that were of any image format, libpng does not, the gbitmap reader end was modified to support bitmapSTB reader, bitmapPNG was reinstated due to a few complications with stb_ not supporting 16bit png files. The length argument is there for 1 point of entry for all format types. Its not used by bitmapPNG or libPNG doesnt inform the reading at all

PR #1670, opened 2026-02-20
Test PR of lazy load of ghosts and scene streaming

marauder2k7:
For now scene streaming only affects how subscenes are loaded. If you try it on everything in a scene expect issues as the namespaces are not linked up at that point.

This is a TEST! Do not merge! If you do you will open up a gate to an alternate reality where all that you believe will cease to exist in an endless void of darkness.


Azaezel:
reminder to extract the lazy instantiation server side from this one
https://github.com/TorqueGameEngines/Torque3D/pull/1670/changes#diff-1390fa94c3a6dd53b3ad702e3762a3c2e5e49fd6a48da1968836caf171f44c18R370
et al

PR #1672, opened 2026-02-21
Expanded Script Assert Context

Areloch:
Caches the processed line in CodeBlock for better assert reporting
Adds additional context on eval and evaluate calls to better isolate script executions that are throwing errors, such as via command fields or eval() calls.

### DIFF
--- a/Engine/source/console/torquescript/codeBlock.cpp
+++ b/Engine/source/console/torquescript/codeBlock.cpp
@@ -35,7 +35,7 @@ using namespace Compiler;
 
 bool           CodeBlock::smInFunction = false;
 CodeBlock *    CodeBlock::smCodeBlockList = NULL;
-StringTableEntry CodeBlock::smCurrentLineText = StringTable->EmptyString();
+const char* CodeBlock::smCurrentLineText = "\0";
 TorqueScriptParser *CodeBlock::smCurrentParser = NULL;
 
 extern FuncVars gEvalFuncVars;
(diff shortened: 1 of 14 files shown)


marauder2k7:
Bug reported by abbey https://discord.com/channels/358091480004558848/783127087820439582/1474514695556825179 

just adding this for reference

Engine/source/console/torquescript/codeBlock.cpp
```diff
@@ -578,6 +579,7 @@ Con::EvalResult CodeBlock::compileExec(StringTableEntry fileName, const char *in
    consoleAllocReset();
 
    name = fileName;
+   smCurrentLineText = inString;
```
dottools, line 582: This pointer should be NULL'd before leaving function.

Engine/source/console/torquescript/codeBlock.h
```diff
@@ -61,6 +61,7 @@ class CodeBlock : Con::Module
 public:
    static bool                      smInFunction;
    static TorqueScriptParser * smCurrentParser;
+   static StringTableEntry smCurrentLineText;
```
dottools: Kind of misleading to use `StringTableEntry` considering each line is not going into the string table. `const char *` should be suffice.

Engine/source/console/torquescript/codeBlock.cpp
Areloch: I went ahead and have it set to "\0" as the initialized value, and then at the two return points in the function which I think should cover the basis. 
Unless you feel using actual NULL would be a bit more specific?

dottools: That does work better since with printf/sprintf() passing a NULL pointer for `%s` is undefined behavior. Some libc implementations did check for NULL and print `(null)` whenever it was NULL, while others like gnu libc will segfault instead.

PR #1688, opened 2026-03-12
Replace 3rd Party builds with VCPKG

marauder2k7:
Replaces all our 3rd party dependencies with vcpkg installed dependencies.
Assimp is still downloaded. 

Jolt physics changes included in this PR


marauder2k7:
outdated just need to do again

PR #1690, opened 2026-03-16
Jolt physics implementation

marauder2k7:
This is linked to the VCPKG pr thats why there are so many changes but this is also a jolt physics initial implementation


marauder2k7:
put into vcpkg pr

PR #1708, opened 2026-04-05
Refactor of sound asset

marauder2k7:
Refactored to match image asset.


marauder2k7:
PR Closed JeffR to do a refactor of all assets simplifying their interfaces and doing away with the macros.

PR #1733, opened 2026-05-11
datatype reworks. vector and tree.

Azaezel:
For the core vector class: Adds the capacity to set a fixed size to a given vector at compile time. Adds the capacity to redirect the standard allocator to a custom one.

New fundamental datatype: TreeNode. structured as a Vector of Vectors of a given type, contains leaf/node logic helpers, as well as a holder example class, TreeObject, which demonstrates serializing to and from the console for any given console-aware data struct. 

Intended usage is for a new generalized backbone for scenegraphs, armatures, nodegraphs, FSM states, ect.

### DIFF
--- a/Engine/source/core/util/tVector.cpp
+++ b/Engine/source/core/util/tVector.cpp
@@ -197,9 +197,12 @@ void test_Vector_dynamic() {
    v.removeIf([](U8 x) { return x == 1; });
    AssertFatal(v.size() == 0, "removeIf failed");
    v.set(arr, 3);
-   AssertFatal(v.countIf([](U8 x) { return x > 1; }) == 2, "countIf failed");
-   AssertFatal(v.anyOf([](U8 x) { return x == 2; }), "anyOf failed");
-   AssertFatal(v.allOf([](U8 x) { return x > 0; }), "allOf failed");
+   U32 countifResult = v.countIf([](U8 x) { return x > 2; });
+   AssertFatal(countifResult, "countIf failed");
+   U8 anyofResult = v.anyOf([](U8 x) { return x == 2; });
+   AssertFatal(anyofResult, "anyOf failed");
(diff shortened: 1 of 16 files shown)


Engine/source/core/util/tVector.cpp
```diff
@@ -99,3 +100,166 @@ bool VectorResize(U32 *aSize, U32 *aCount, void **arrayPtr, U32 newCount, U32 el
 }
 
 #endif
+
+void test_Vector_dynamic() {
```
marauder2k7, line 104: do ya want this broken out to a unit test ?

marauder2k7 (approved):

Engine/source/core/util/tVector.cpp
Azaezel: sure

Engine/source/core/util/treeObject.cpp
```diff
@@ -0,0 +1,472 @@
+//-----------------------------------------------------------------------------
```
jamesu, line 1: Does GarageGames have perpetual 2012 copyright on these new files? Seems a bit bizarre!

Azaezel: Fair. could always put "portions copyright by" if folks would prefer,  since most of it is a variation on existing architecture.

jamesu: Its more a legal concern rather than a community focussed one; if we were to presume this code is a new work (which it STRONGLY looks like), right now the header attributes the wrong year for copyright. If the author wasn't actually GarageGames, this may contribute to problems later on as the work is mis-attributed to the wrong author.
Additionally, it would be slightly odd for an author to be applying license conditions from a different file (i.e. including the outdated copyright notice) in a brand new work.

PR #1736, opened 2026-05-13
Update gfxGLCircularVolatileBuffer.h

marauder2k7:
should fix the volatile buffer issues with opengl


marauder2k7:
PR worked for 1 use case, but completely failed and made things worse on another machine. Possible issues with buffer storage support vs not having it, its the only way the code diverges......

PR #1743, opened 2026-05-17
Jolt Physics implementation without vcpkg

marauder2k7:


dottools:
I'm confused. Wasn't many dependencies moved from local source copy to vcpkg recently and now jolt is being moved to a local source copy? Why?

marauder2k7:
> I'm confused. Wasn't many dependencies moved from local source copy to vcpkg recently and now jolt is being moved to a local source copy? Why?

Its not agreed yet whether or not we will be switching to vcpkg so needed to do an implementation without vcpkg just to be ready for whichever way is decided

Azaezel:
The general philosophy is not to rely on assuming dependencies are just always going to be in a verified, tested state hosted "someplace in the cloud". No matter what indirect tooling that is used to accomplish that. For some cases there's no getting around it. But we'd like to keep that to a minimum.

marauder2k7:
yeah atm the only one restricted to being downloaded is assimp because the repo doesnt support lfs, although granted, that will only happen once no matter how many projects you create etc but still

PR #1745, opened 2026-05-18
Proper constructor for PathShapeData

Areloch:
Has the PathShapeData constructor proper initialize the mUseEase var which prevents the datablock CRC cache from failing to match.


TristonC13:
question: why write a new constructor to assign the value of false to a variable when you can just initialize it to be false in the class definition?

PR #1754, opened 2026-05-31
ShapeAsset Cleanup and Standardization refactor

Areloch:
- Cleans up ShapeAsset of some unnecessary/redundant elements like extra material and animations tracking
- Removed the old SHAPE_ASSET macros
- Implements AssetRef struct that acts as a universal wrapper for an templated AssetPtr and AssetId pair Adds Type handling for AssetRef for ShapeAsset to unify handling in classes that utilize a shapeAsset, so assigning an assetPtr or an assetId will keep a record of the assignment in the event the assetPtr is invalid. 
- Update all classes that utilized the old SHAPE_ASSET macros to utilize the AssetRef struct and updated the class code to utilize it to provide much more clean and concise code that isn't blocked behind macro definitions 
- Added a new example class: shapeDatablockExample which allows render of a simple shape object utilizing a simple example datablock.

### DIFF
--- a/Engine/source/T3D/assets/ShapeAsset.cpp
+++ b/Engine/source/T3D/assets/ShapeAsset.cpp
@@ -55,15 +55,11 @@
 #include "gfx/bitmap/imageUtils.h"
 
 StringTableEntry ShapeAsset::smNoShapeAssetFallback = NULL;
+AssetPtr<ShapeAsset> ShapeAsset::smNoShapeAssetFallbackAssetPtr = NULL;
 
 //-----------------------------------------------------------------------------
 
 IMPLEMENT_CONOBJECT(ShapeAsset);
-
-
-//-----------------------------------------------------------------------------
-// REFACTOR
-//-----------------------------------------------------------------------------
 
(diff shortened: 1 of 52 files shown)


Engine/source/T3D/assets/ShapeAsset.cpp
```diff
@@ -131,8 +127,50 @@ ConsoleSetType(TypeShapeAssetId)
 }
 
 //-----------------------------------------------------------------------------
-// REFACTOR END
-//-----------------------------------------------------------------------------
+IMPLEMENT_STRUCT(AssetRef<ShapeAsset>, AssetRefShapeAsset, , "")
+END_IMPLEMENT_STRUCT
+
+ConsoleType(ShapeAssetRef, TypeShapeAssetRef, AssetRef<ShapeAsset>, ASSET_ID_FIELD_PREFIX)
```
marauder2k7, line 133: Does the taml end know about AssetRef<ShapeAsset> as a type?

Areloch: Yep testing shows it parses it into the dependencies lists when the taml is scanned as expected

PR #1770, opened 2026-06-17
Test for whether configure depends will help with the precompiled header issue

marauder2k7:
The issue with adding source and header files to the project while using cmakes precompiled header may have been caused by an improper glob setup. Configure depends ensures that cmake and the generated project stay in sync with one another.


marauder2k7:
produced new issues for steve 

https://discord.com/channels/358091480004558848/416629987936829451/1516766577675079761

PR #1784, opened 2026-07-08
remove redundant text entries

Azaezel:
for GuiTextEditCtrl elements, the text *is* the variable. so filling that entry out overwrites the variable itself. just use obj.variable = $pref::foo; patterns there to ensure lookups. for server config purposes, added an exec("core/clientServer/scripts/server/defaults"); to keep those config values distinctly clustered while ensuring the same order of operations as other prefs.


Azaezel:
Issue reported, and reported as resolved with this PR by Steve_Yorkshire: https://discord.com/channels/358091480004558848/580242040730943509/1524442046226632994

PR #1792, opened 2026-07-17
same fixes for neon

marauder2k7:
Performance tests on neon showed the same as the other intrinsics.

The work for all the intrinsics is there just need to go through it with a skeptical eye and see how we can actually increase performance instead of damage it.

### DIFF
--- a/.github/workflows/build-windows-msvc.yml
+++ b/.github/workflows/build-windows-msvc.yml
@@ -22,18 +22,6 @@ jobs:
             matrix:
                 config:
                     - {
-                          name: "Windows MSVC Ninja",
-                          os: "windows-latest",
-                          build_type: "Release",
-                          cc: "cl",
-                          cxx: "cl",
-                          generator: "Ninja",
-                          cmake_args: "",
-                          compiler_family: "msvc",
-                          msvc_arch: "amd64",
-                          can_test: false,
-                      }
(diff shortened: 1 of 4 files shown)


dottools:
FYI the industry practice to disable blocks of code is via `#if` macro, not block comments, like so:

```cpp
#if 0
// ..
// code
// ..
#endif // 0
```

Reason is that `#if`'d out code shows as disabled, versus commented code just looks like any other comments, in typical code editors / IDEs. Also looks cleaner on diff's to show intention.

marauder2k7:
> FYI the industry practice to disable blocks of code is via `#if` macro, not block comments, like so:
> 
> ```c++
> #if 0
> // ..
> // code
> // ..
> #endif // 0
> ```
> 
> Reason is that `#if`'d out code shows as disabled, versus commented code just looks like any other comments, in typical code editors / IDEs. Also looks cleaner on diff's to show intention.

i get that but as it stands at the moment this will be the same as the other isa expansions, they all need some more vigorous tests and usually these if defs can be switched in in chinks, we would need one for each function ptr being installed if that is the case because this is being tested against other implementations and if 1 of these functions after changes becomes more performant than the scalar it will be uncommented out. 

Active tests are being run this was just a quick kill all until its figured out

PR #1794, opened 2026-07-20
legacy DAE filenme support

Azaezel:
more thana  few tools use capital DAE, so we need to ensure the lookup there is non case sensitive


Azaezel:
@marauder2k7 has an alternative fix incomming

